PR #82, opened 2014-09-12
made Box behave like h2d.comp.Interactive

darkdarkdragon:
Proposition for adding more interactivity to Box
Here I just copied some from h2d.comp.Interactive
It has two problems: 

1. copy-paste and fact that scripts bound 
2. h2d.comp.Parser binds scripts to properties like 'onclick' only to h2d.comp.Interactive

Another way to do this is to just sub-class Box form Interactive

So what would be preferred way?


ncannasse:
h2d.comp is deprecated

PR #99, opened 2015-03-16
Reference documentation: Displaying Text

Namide:
cleans tags ;
describes h2d.Text ;
describes h2d.Font ;

### DIFF
--- a/doc/ref.html
+++ b/doc/ref.html
@@ -39,9 +39,9 @@
 </p>
 
 <ul>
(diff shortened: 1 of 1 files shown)


ncannasse:
Could you update the PR by keeping tabs instead of multiple spaces ? thanks :)

Namide:
Yes :)
I hadn't seen that I was in spaces mode.

PR #107, opened 2015-08-03
Ignore files/directories starting with . or _

jonongjs:
`FileTree.scanRec` currently only ignores directory names beginning with `'.'` or `'_'`. It should also ignore file names beginning with those characters. (It crashes when encountering file names beginning with `'.'`)

### DIFF
--- a/hxd/res/FileTree.hx
+++ b/hxd/res/FileTree.hx
@@ -250,9 +250,9 @@ class FileTree {
 			var fileName = f;
 			var field = null;
 			var ext = null;
+			if( f.charCodeAt(0) == ".".code || f.charCodeAt(0) == "_".code )
+				continue;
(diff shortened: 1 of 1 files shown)


ncannasse:
I actually noticed we're using this when we have a file such as "8bit.png" , what kind of error are you getting ?

jonongjs:
I get an `Invalid_argument("index out of bounds")` error at compile time when a dotfile exists in the res directory. This is due to `FileTree.scanRec` trying to add a field with an empty identifier. On the other hand, I have no issues with a file such as "8bit.png". There shouldn't be a problem, I believe, since, for files, only those starting with "." or "_" are ignored.

ncannasse:
Could you open an issue for this ? I'll fix it, your patch prevents me from accessing  hxd.Res._8bit ressource

Issue #115, opened 2015-09-18
GLSL compilation with helper functions


ncannasse:
ATM it is not possible to compile GLSL code for shaders which contains helper functions (sample/SAO)

ncannasse:
Because of the lighting shaders change in https://github.com/ncannasse/heaps/commit/da87ea910fd35a479e016b28495442656612403a this breaks all GL samples, will try to fix it asap

delahee:
Thanks !

2015-11-18 20:29 GMT+01:00 Nicolas Cannasse <notifications@github.com>:

> Closed #115 <https://github.com/ncannasse/heaps/issues/115> via 088249e
> <https://github.com/ncannasse/heaps/commit/088249e2c00dafb6532fc83951d007b1b7de276e>
> .
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ncannasse/heaps/issues/115#event-468350604>.
>



-- 
David Elahee

PR #123, opened 2015-10-20
Add support for openfl/lime

mrcdk:
I've only tested it with this sample https://github.com/ncannasse/heaps/tree/master/samples/basic

The `hxml` to get it to work is:
```
-cpp bin
-main Main
-lib heaps
-lib openfl
-lib lime

-D lime-openal
-D openfl-next
-D lime-cairo
-D lime-opengl
-D native
-D lime-native
-D lime-curl
-D openfl-native
-D windows
-D desktop


-D resourcesPath=../res
--macro include('h3d')
--remap flash:openfl
```
And I had to comment out these lines for it to run:
https://github.com/ncannasse/heaps/blob/master/h3d/shader/AmbientLight.hx#L31
https://github.com/ncannasse/heaps/blob/master/h3d/shader/AmbientLight.hx#L35
https://github.com/ncannasse/heaps/blob/master/hxsl/Linker.hx#L365

### DIFF
--- a/h3d/impl/Driver.hx
+++ b/h3d/impl/Driver.hx
@@ -12,6 +12,10 @@ typedef Texture = { t : js.html.webgl.Texture, width : Int, height : Int, fmt : 
 typedef IndexBuffer = nme.gl.GLBuffer;
 typedef VertexBuffer = { b : nme.gl.GLBuffer, stride : Int };
 typedef Texture = { t : nme.gl.GLTexture, width : Int, height : Int, fmt : Int, ?fb : nme.gl.GLFramebuffer, ?rb : nme.gl.GLRenderbuffer };
+#elseif lime
+typedef IndexBuffer = lime.graphics.opengl.GLBuffer;
+typedef VertexBuffer = { b : lime.graphics.opengl.GLBuffer, stride : Int };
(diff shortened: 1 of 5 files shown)


ncannasse:
Thanks !
The three commented lines are related to #115, will be fixed soon

kevinresol:
Does this mean openfl native extensions can be used with heaps?

mrcdk:
@kevinresol I don't think so, not without the lime build system. I haven't tried to use it with heaps and see how ~~my computer explodes~~ they work together. The lime build system is also used in some targets (android, iOS) to create the final executable, so getting it to work with heaps is needed. I'll try it soon.

It may be possible to remove openfl from the equation altogether... I will investigate this too.

delahee:
Well it should, our workflow at MT was openfl native extension + h3d which
is pretty much the same as heaps so yes it can definitely work and used to
in our case.

2015-10-21 19:40 GMT+02:00 Justo Delgado <notifications@github.com>:

> @kevinresol <https://github.com/kevinresol> I don't think so, not without
> the lime build system. I haven't tried to use it with heaps and see how my
> computer explodes they work together. The lime build system is also used
> in some targets (android, iOS) to create the final executable, so getting
> it to work with heaps is needed. I'll try it soon.
>
> It may be possible to remove openfl from the equation altogether... I will
> investigate this too.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ncannasse/heaps/pull/123#issuecomment-149973272>.
>



-- 
David Elahee

PR #142, opened 2016-01-15
Fix textWidth with align Center, Right

Tipyx:


ncannasse:
Could you resubmit with clean changes (no newline/spaces changes) ?
Also, I don't understand the change: calcWidth needs to be written whatever the alignment

Tipyx:
Yep, but when you update the text with align Center or Right, calcWidth = realTextWidth + x (which is needed by getBoundsRec), it's why I separate the values in two differents parameters :
- calcWidth : the real size of the text (regardless of the Align)
- calcXMax : calcWidth + offsetX

ncannasse:
I see. You still need to save/restore calcXMax in calcTextWidth

Tipyx:
It's done.

ncannasse:
This breaks h2d.Flow, I have no idea why so I reverted for now

PR #146, opened 2016-02-22
Update GlslOut.hx

kiroukou:
specify the shader version. 
1.3 being the lower version which allow precision qualifier  (glsl defaut version being 1.1)

### DIFF
--- a/hxsl/GlslOut.hx
+++ b/hxsl/GlslOut.hx
@@ -372,6 +372,7 @@ class GlslOut {
 		decls = [];
 		buf = new StringBuf();
 		exprValues = [];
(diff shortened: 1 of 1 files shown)


ncannasse:
Could you check if it runs with WebGL ?
http://stackoverflow.com/questions/21103484/is-it-possible-to-run-version-120-shaders-with-webgl

kiroukou:
cleaned a bit since pull request didn't came from the repo I expected (personal and not motiontwin)

Concerning the precision qualifier on WebGL, it is not supported indeed.
I'll make an other change and test it.

PR #147, opened 2016-02-22
Lime/Openfl better integration

kiroukou:
Small changes for a working openfl/lime support.

The one that might look wierd is : flash.ui.Keyboard.BACK thing, but openfl has named that key BACKSPACE....
hence the keycode instead.

also, openfl being used in parallel with flash, some changes to conditional conds are made.


ncannasse:
BACK != BACKSPACE : Back is the Android Back key :)

kiroukou:
Ok my mistake.

We are currently moving that pull request forward.
Our goal is to have a robust lime integration  (no openfl as much as possible, to avoid another dependency).

I'll let you know when the pull request is complete now that ppedidont did some other changes.

kiroukou:
Current branch does have a correct lime integration.
Note : 
 * Lime CPP tested on Windows, Linux, IOs and Android  (more info below concerning JS)
 * Assets:
Lime Assets integrated with hxd.Res system using hxd.Res.initLime
Lime xml project format should however use forced binary format for its assets:
Example : <assets path="assets" embed="false" rename="" include="*.jpg|*.png" type="BINARY" />

 * Events / (Multi) Touch / Keys
 * FontBuilder


Missing:
 * JS target is currently not working properly.  We do not use that target so far, so we will not investigate a lot, but it compiles correctly.
(Error being : Unable to get property 'pixelStorei' of undefined or null reference     would need to check DCE or compilation flags)

ncannasse:
JS target does require a `<canvas id="webgl"/>` element on the webpage, but it should error if not found. If you look at h3d/imp/GlDriver.hx I don't see how this could happen.

ncannasse:
Regarding the PR : everything seems fine, except the changes h2d.Text : I don't want to use the haxe.Utf8 API because HL uses native UCS2 text (same as JS) which is not the same as UTF-8. I guess we need to define an heaps-specific abstract to handle that for now, until Haxe has proper std Unicode support.

pperidont:
I have removed calls to haxe.Utf8 from h2d.Text and created a new abstract in hxd.

ncannasse:
Do you still need the fixMult part ? It should have been fixed for a quite long time already.

ncannasse:
Nevermind, it was simply moved code.

ncannasse:
Thanks!

PR #148, opened 2016-02-22
Update GlslOut.hx

kiroukou:
shader version support on desktop only for being abel to use the precision qualifier


kiroukou:
wasn't able to confirm the change on sample/basic/ogl  since I still have the crash (with&without the change), but shouldn't be a problem.

PR #149, opened 2016-03-01
Update GlslOut.hx

kiroukou:
Mobile using openglES were also concerned by this.

It would be certainly better to detect opengles in some way, rather than using such conditionals, but so far it works.

### DIFF
--- a/hxsl/GlslOut.hx
+++ b/hxsl/GlslOut.hx
@@ -373,7 +373,7 @@ class GlslOut {
 		buf = new StringBuf();
 		exprValues = [];
 		//Version is required on desktop for precision qualifier, but version 1.0.0 only is supported on webgl
-		#if !js
+		#if (windows||linux)
 		decls.push("#version 130");
 		#end
 		decls.push("precision mediump float;");


ncannasse:
I'm not sure that's a good flag to be used in these cases. I guess all mobile run ES, so maybe #if (cpp && mobile) ?

delahee:
Hi guys, my 2c, be aware some desktop targets ( consoles and webgl for
example ) use opengles-GLSL too. So I suggest you explicitly discriminate
gles with gldesktop..

2016-03-01 12:41 GMT+01:00 Nicolas Cannasse <notifications@github.com>:

> I'm not sure that's a good flag to be used in these cases. I guess all
> mobile run ES, so maybe #if (cpp && mobile) ?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ncannasse/heaps/pull/149#issuecomment-190683470>.
>



-- 
David Elahee

kiroukou:
Well this should be set only on desktop apps afaik.
JS and mobile should not define version number.

I agree that is not the perfect solution, being able to define correctly like david said a gles flag would be of course better.

ncannasse:
Any update on this PR?

ncannasse:
I'm closing, please submit again another time :)

kiroukou:
Ok. I'm currently working in this.
I thnk i have a working version of runtime shader setup.
Le 27 avr. 2016 16:08, "Nicolas Cannasse" <notifications@github.com> a
écrit :

PR #158, opened 2016-03-21
Fix alloc not occurring until render

sepharoth213:
This fixes a bug where if the points in a `Polygon` primitive are edited before the first render but after a mesh is created with that primitive, then the mesh display the new points. Changing the points any time after that will not affect the object, so this behavior should be more logically consistent.

### DIFF
--- a/h3d/scene/Mesh.hx
+++ b/h3d/scene/Mesh.hx
@@ -10,6 +10,7 @@ class Mesh extends Object {
 		this.primitive = prim;
 		if( mat == null ) mat = new h3d.mat.MeshMaterial(null);
 		this.material = mat;
+		this.primitive.alloc(null);
 	}
 
 	override function getBounds( ?b : h3d.col.Bounds, rec = false ) {


ncannasse:
I think that the behavior should be instead that the Mesh always display the new points.

sepharoth213:
I agree. That would require calling `alloc()` on the primitive whenever the points are changed, but that would have to be handled on a by-primitive basis.

Should each primitive handle reconstructing the buffer on changes or should the `Primitive` or `Mesh` class handle this somehow? A flag in the `Primitive` class that the user sets like `prim.needUpdate = true` might be a better solution: then the primitive would `alloc()` when it comes time to render.

ncannasse:
A primitive with a null or disposed buffer will call alloc() at render time already, so we only need to call dispose() when the points are changed.

sepharoth213:
So primitives, specifically Polygon, can:
 - Use a custom Array class that has callbacks for when its members are changed or
 - Cache a deep copy of each array that needs to be checked and compare them each render or
 - Leave it up to the user to call dispose() when they make any changes

I think the second option is the least impactful on the API and implementation-wise, but performance-wise it might not be the greatest solution.

ncannasse:
I think the third option is the best. You can submit appropriate documentation for it if you want.

PR #161, opened 2016-03-24
Timer improvements

sepharoth213:
I noticed that the timer was making my timers inaccurate with respect to real time, so I added an option for turning the `deltaT` smoothing off.

I also fixed a bug where if `deltaT` was greater than `maxDeltaTime`, it would fall back to the previous frame's `calc_tmod`.

### DIFF
--- a/hxd/Timer.hx
+++ b/hxd/Timer.hx
@@ -9,6 +9,9 @@ class Timer {
 	public static var calc_tmod : Float = 1;
 	public static var tmod : Float = 1;
 	public static var deltaT : Float = 1;
+
+	public static var smoothing : Bool = true;
+
 	static var frameCount = 0;
 
 	public inline static function update() {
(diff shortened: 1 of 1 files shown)


ncannasse:
I think you can already disable smoothing by setting tmod_factor=0

sepharoth213:
Ah, that's what the code was intended to do. I've changed it so that setting `tmod_factor` to 0, it will behave as if there was no smoothing. For some reason, heaps had previously had `(1 - tmod_factor) * deltaT * wantedFPS`, making the reported dt really broken. That's where the issues I was having with timing were coming from.

I also fixed fps() reporting `wantedFPS` times the current framerate rather than the actual current framerate.

ncannasse:
I don't want to change calc_tmod at all, the idea of the tmod is that is should = 1 when your actual FPS  match the wantedFPS.

sepharoth213:
I was confused because App calls update with tmod passed in as dt: `update(hxd.Timer.tmod);`
I assumed that meant it was supposed to be some other form of deltaT.

Is that intended as well?

ncannasse:
Yes

PR #164, opened 2016-03-25
Added containsXY method (allocation-free check)

deepnight:
This is soooo cool, you really should accept it.


ncannasse:
Because of inline of both contains and Point contructor, you can safely do  .contains(new Point(5,6)) and it will remove the allocation.
Also, you got a isGlow commit mixed in the PR.

PR #166, opened 2016-04-04
Console hide, show and fix log

hammeron-art:
Let hide and show by code.
The log wasn't working with Haxe 3.2.1 because the UString class and this: https://github.com/HaxeFoundation/haxe/issues/4426

### DIFF
--- a/h2d/Console.hx
+++ b/h2d/Console.hx
@@ -227,7 +227,23 @@ class Console extends h2d.Sprite {
 		logs.push(command);
 		logIndex = -1;
 
-		var args = ~/[ \t]+/g.split(command);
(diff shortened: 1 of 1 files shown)


ncannasse:
What's the regexp for ? It seems to somehow restrict arguments values, not sure I want it :)

hammeron-art:
It doesn't restrict arguments.
I made it to match regular arguments and strings with spaces. http://regexr.com/3d58g
I put some extra things like (?:['|"]) to match but don't capture quotes but Haxe regex don't expose functions to deal with capture groups.

ncannasse:
The regexp cause some problem, for instance if you do `/command myfile.png` it will split in two `myfile` and `png`

hammeron-art:
Yeah, can't use regex at all for this.
Now with a loop it works for symbols too.

PR #175, opened 2016-07-20
Fixed a bug in SpriteBatch bounds calculation

Stee92:
Bounds for a non scaled / rotated SpriteBatch only got calculated with the tile that is used to initialize the SpriteBatch.
Instead the tiles of the batch elements should be used to calculate the bounds. ;)

I do not like the fact that we have to pass a Tile to create a SpriteBatch. Instead we could pass the texture / atlas? It is a bit confusing when you first look at how it's done. Suggestions?


ncannasse:
Thanks for the patch.
Regarding using a Tile instead of a Texture, I choose to do so for SpriteBatch and TileGroup because h2d API mostly use Tile instead of Texture.

PR #178, opened 2016-08-05
TileGroup : addTransform function

mcapdegelle:


ncannasse:
I think you don't want to use  toColor + insertColor that does a lot of useless Float -> Int -> Float conversions

PR #189, opened 2016-10-10
ZGroup

mcapdegelle:
Add a ZGroup wrapper.
Renders children in two passes to take advantage of z-culling.

- Opaque drawables are rendered front-to-back and write into the z-buffer
- Semi-transparent drawables are rendered back-to-front and read the z-buffer

Set "blendMode" to "None" on opaque drawables whenever you can.
If a drawable only contains fully-opaque and fully-transparent pixels, its blendmode may be set to "None" since alpha kill is used in the opaque pass.

### DIFF
--- a/h2d/RenderContext.hx
+++ b/h2d/RenderContext.hx
@@ -10,6 +10,12 @@ class RenderContext extends h3d.impl.RenderContext {
 	public var textures : h3d.impl.TextureCache;
 	public var scene : h2d.Scene;
 	public var defaultFilter : Bool = false;
+	public var killAlpha : Bool;
+	public var front2back : Bool;
+
+	public var onBeginDraw : h2d.Drawable->Bool; // return false to cancel drawing
+	public var onEnterFilter : h2d.Sprite->Bool;
+	public var onLeaveFilter : h2d.Sprite->Void;
 
 	public var tmpBounds = new h2d.col.Bounds();
 	var texture : h3d.mat.Texture;
@@ -25,6 +31,7 @@ class RenderContext extends h3d.impl.RenderContext {
 	var stride : Int;
 	var targetsStack : Array<{ t : h3d.mat.Texture, x : Int, y : Int, w : Int, h : Int, renderZone : {x:Float,y:Float,w:Float,h:Float} }>;
 	var hasUVPos : Bool;
+	var filterStack : Array<h2d.Sprite>;
 	var inFilter : Sprite;
 
 	var curX : Int;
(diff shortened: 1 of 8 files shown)


h2d/RenderContext.hx
```diff
@@ -212,24 +238,25 @@ class RenderContext extends h3d.impl.RenderContext {
...
 		baseShader.absoluteMatrixB.set(obj.matB, obj.matD, obj.absY);
 		beforeDraw();
+		return true;
 	}
 
 	@:access(h2d.Drawable)
 	public function beginDrawBatch( obj : h2d.Drawable, texture : h3d.mat.Texture ) {
-		beginDraw(obj, texture, false);
+		if( !beginDraw(obj, texture, false) ) return false;
+		return true;
```
ncannasse: return beingDraw ?

h2d/RenderContext.hx
```diff
@@ -249,14 +276,16 @@ class RenderContext extends h3d.impl.RenderContext {
 			absY = obj.absY;
 		}
 
+		if( !beginDraw(obj, tile.getTexture(), true, true) ) return false;
```
ncannasse: We do not want to move this before checking out of screen, because this might cause shader switching for out of screen elements (move it back after the check, does it change something?)

h2d/Sprite.hx
```diff
@@ -386,10 +386,10 @@ class Sprite {
 			nullDrawable.matB = matB;
 			nullDrawable.matC = matC;
 			nullDrawable.matD = matD;
-			ctx.drawTile(nullDrawable, tile);
+			if( !ctx.drawTile(nullDrawable, tile) ) return;
```
ncannasse: Not really necessary

h2d/RenderContext.hx
```diff
@@ -212,23 +238,25 @@ class RenderContext extends h3d.impl.RenderContext {
...
 
 	@:access(h2d.Drawable)
 	public function beginDrawBatch( obj : h2d.Drawable, texture : h3d.mat.Texture ) {
-		beginDraw(obj, texture, false);
+		return beginDraw(obj, texture, false);
 	}
 
 	@:access(h2d.Drawable)
 	public function drawTile( obj : h2d.Drawable, tile : h2d.Tile ) {
+		if( !beginDraw(obj, tile.getTexture(), true, true) ) return false;
```
ncannasse: Here too we are moving the beginDraw (potential shader switch) before the bounds check, which will reduce performances for outside of screen element.

h2d/RenderContext.hx
```diff
@@ -212,23 +238,25 @@ class RenderContext extends h3d.impl.RenderContext {
...
```
ncannasse: Here too we are moving the beginDraw (potential shader switch) before the bounds check, which will reduce performances for outside of screen element.

PR #197, opened 2016-11-04
Stencil

mcapdegelle:
Add support for stencil buffer in materials


ncannasse:
Looks good, but I think we should group the stencil values into a different object than Pass, so you would access it with `pass.stencil.***` and it would be null by default (so 0 to all bits values)
Can you make the necessary changes? TY

ncannasse:
Why is there need for stencil bits in material? They should only be in the stencil object

pperidont:
up :) (conflict resolved)

PR #207, opened 2016-12-27
fix Anim not work when set frames in a later time.

LouisSong:
create the Anim instance,then set the frames after atlas source loaded in a later time,the frames.length equals 0,"curFrame %= frames.length" make curFrame to NaN,so the anim stoped on the first  
 frame.


ncannasse:
I think in that case the fix should be if( frames.length == 0 ) curFrame = 0 else ....

LouisSong:
I have changed the code as what you said.

PR #214, opened 2017-01-25
Add support for local res watch() on sys platforms

pperidont:


ncannasse:
Looks ok, maybe doing the same check as AIR3 to see if the file is currently being written would also allow to prevent change event to be triggered while the file is still being written by some 3rd party app ?

PR #230, opened 2017-02-13
Textinput working on OSX Chrome

thomasuster:
Textinput wasn't working at all on OSX chrome, this fixed it. Also added some keycodes for to prevent fall through character to be added if pressing left or right command button or shift.


ncannasse:
Sadly this is not correct, since depending on the browser either charCode or keyCode will make sense. I'm planning to have a separate event, see #231

thomasuster:
Cool, thanks for opening the issue.

Issue #144, opened 2016-02-09
Bug on Interactive


ncannasse:
In the `interact` sample, having both rotation and scaleX on h2d.Interactive will not work correctly.
Most likely related to Scene.screenToLocal not behaving as it should.

ncannasse:
That's still an issue

ncannasse:
Negative scale is also not working

PR #281, opened 2017-09-14
Take rotation into account for mouseX/Y and event handling in Scene (fixes #51 and #144)

### LINKED ISSUE #51: h2d.Scene rotation/events
Scaling the h2d.Scene is not properly supported atm. This needs to be handled (requires fixing event handling and mouseX/Y as well)

trethaller:
Not sure if `globalToLocal` was originally avoided for performance reasons, but this seems to work well.
I've also modified the `Interactive` sample to test combinations of scene rotation and interactive rotation + scaling, as well as testing the `getInteractive` function.

### DIFF
--- a/h2d/Scene.hx
+++ b/h2d/Scene.hx
@@ -79,11 +79,17 @@ class Scene extends Layers implements h3d.IDrawable implements hxd.SceneEvents.I
 	}
 
 	function get_mouseX() {
-		return this.globalToLocal(new Point(stage.mouseX, stage.mouseY)).x;
+		var mx = stage.mouseX - absX;
+		var my = stage.mouseY - absY;
+		var invDet = 1 / (matA * matD - matB * matC);
+		return (mx * matD - my * matC) * invDet;
 	}
 
 	function get_mouseY() {
-		return this.globalToLocal(new Point(stage.mouseX, stage.mouseY)).y;
+		var mx = stage.mouseX - absX;
+		var my = stage.mouseY - absY;
+		var invDet = 1 / (matA * matD - matB * matC);
+		return (-mx * matB + my * matA) * invDet;
 	}
 
 	public function dispatchListeners( event : hxd.Event ) {
--- a/h2d/Scene.hx
+++ b/h2d/Scene.hx
@@ -1,4 +1,5 @@
 package h2d;
+import h2d.col.Point;
 import hxd.Math;
 
 class Scene extends Layers implements h3d.IDrawable implements hxd.SceneEvents.InteractiveScene {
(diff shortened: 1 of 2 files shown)


ncannasse:
Thanks! I'm in Amsterdam till Saturday, will look at it asap.

trethaller:
Just thinking that in most cases scene won't have a transform, maybe it's worth optimizing for these cases

ncannasse:
Maybe it would be better avoiding globalToLocal/localToGlobal since they make syncPos() which we are supposed to have already handled at this point, and that would require looking through the whole parents of interactive which is better if we can avoid.

And yes optimizing for no transform can be a good thing.

trethaller:
Ok I'll avoid globalToLocal. More questions/comments:

1. Can you confirm that Scene.getInteractive takes x,y params that are local to the scene whereas events passed to handleEvent/dispatchEvent have their coordinates (relX, relY) in Stage space?
2. The current version on master applies the transformation twice in handleEvent/dispatchEvent (first does screenToLocal, then uses the absolute transform of the Interactive) so it also doesn't work when a translation and scaleX is applied to s2d in the Interactive sample (without rotation). I'm also fixing that

trethaller:
Just found out that I was not handling setFixedSize properly, I'll look into that as well.

ncannasse:
@trethaller yes, for (1) ,  and thanks for (2) I might have not tested everything as I should have regarding transforms being applied to the scene itself

trethaller:
Okay I think I have everything working:

- dispatchEvent
- handleEvent
- getInteractive
- mouseX/mouseX

All support scaling, fixedSize, rotation, and x/y offset applied to Scene (all tested in the Interactive sample).

I've also optimized mouseX/mouseY for the case where rotation is 0. I've tried to avoid repetition when possible, that's be best I could come up with without doing unnecessary calculations.

ncannasse:
`toInteractiveLocal` performs an allocation of a GC value h2d.col.Point, which should be avoided for such calculus. You can try to add inline to it + compile with -D dump=pretty to output the optimized code of h2d.Scene and check that the allocation has been correctly optimized out.

trethaller:
Yes, with inline the allocation goes away:

		var local_y;
		var local_x;
		var px = event.relX - i.absX;
		var py = event.relY - i.absY;
		var invDet = 1 / (i.matA * i.matD - i.matB * i.matC);
		var lx = (px * i.matD - py * i.matC) * invDet;
		var ly = (- px * i.matB + py * i.matA) * invDet;
		local_x = lx;
		local_y = ly;

I'll commit that

trethaller:
Didn't know this worked btw that's cool :-)

ncannasse:
It is due to h2d.col.Point constructor being marked as inline ;)

ncannasse:
We spotted this causes a regression in Northgard in some specific UI parts (loading/saving games and options menu). We will try to submit a reproducible example.

ncannasse:
PS: only the X coordinates is affected, the interactive zone is much more on the right than it should.

trethaller:
Oops, sorry about that :(
Could it have to do with there being a hierarchy? This is not something that I've tested in the Interactive sample.

ncannasse:
Not a problem, it didn't made it live. We couldn't really figure out what were the exact reproducing condition but there's for sure something weird going on in that specific case. We have reverted for now, I guess you'll have to look at it again in a few weeks ;)

PR #284, opened 2017-09-23
default value for textAlign if it's set to null

josuigoa:


ncannasse:
We don't allow setting null values for many different enums so I don't want to make an exception here. You're not supposed to assign "null" to it, that's how the API works.

josuigoa:
I got an error that I solved adding that line. I thought that there was a strange behaviour where align can get null value, but I've seen it is my code's fault... :sweat_smile:

PR #289, opened 2017-10-10
Support for pad on hldx 

pperidont:


ncannasse:
- rumble time should be in seconds, not ms (heaps API only uses seconds)
- handling the delays in syncPads instead of using haxe.Timer would be nicer
- it seems strange that CONFIG_DX is used in syncPads...

PR #291, opened 2017-10-10
Misc Fixes 

pperidont:
- Avoid texture creation in `CachedBitmap` if width or height is zero (fails on DirectX)
- Call flushTarget() before driver.setRenderZone() (`GlDriver` uses `curTarget.height` in setRenderZone)
- Check `FileSystem.exists` before `isDirectory` (isDirectory throws an exception if destination is not accessible)
- Other minor DirectX fixes


ncannasse:
I wonder about the pixelAlign thing. I thought it was no longer necessary in DX10+

ncannasse:
If you remove pixelAlign I can merge the rest

PR #300, opened 2017-10-29
add Spotlight shader sample

Simn:
The shader code is ported and simplified from https://www.tomdalling.com/blog/modern-opengl/08-even-more-lighting-directional-lights-spotlights-multiple-lights/

I left out some aspects (specular etc.) to focus on how to actually write shaders.

### DIFF
--- a/samples/SpotlightSample.hx
+++ b/samples/SpotlightSample.hx
@@ -18,7 +18,7 @@ class SpotlightShader extends hxsl.Shader {
 			surfaceToLight = normalize(surfaceToLight);
 			// Calculate the angle between inverse surfaceToLight vector (= pointing from light to the pixel)
 			// and the light cone angle vector.
-			var lightToSurfaceAngle = acos(dot(-surfaceToLight, normalize(lightConeDirection)));
+			var lightToSurfaceAngle = dot(-surfaceToLight, lightConeDirection);
 			var attenuation:Float;
 			// If the calculated angle is larger than the light cone angle, the pixel is not affected by the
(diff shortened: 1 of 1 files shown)


ncannasse:
I don't think it should be necessary to calculate the acos of the angle, one method only involving vector dot product should be feasible.

ncannasse:
Also, lightConeDirection should be normalized before being passed to the shader, so you don't have to renormalize for every pixel

Simn:
I'll fix the acos one.

For the direction, I think at the very least I have to clone the vector because having x/y/z sliders and normalizing doesn't mix that well.

Simn:
Changed both, please check.

My documentation is off now because the comparison operator flipped (`lightToSurfaceAngle < lightConeAngle`). Have to understand why exactly before I update the documentation...

ncannasse:
Another optimization :  pow( length(surfaceToLight) , 2.0)  is actually surfaceToLight.dot(surfaceToLight)

Simn:
Whoops, I meant to get rid of that clone entirely. But yes, loading into the shader variable makes even more sense.

ncannasse:
Regarding your change : I don't have previous code but `a.dot(b) == a.length() * b.length() * cos(a,b)`

Simn:
Updated the direction and the pow. Still have to check the cos.

ncannasse:
Closing for now, please resubmit not as a sample but as separate files (h3d/SpotLight.hx , h3d/shader/SpotLight.hx + sample).

PR #311, opened 2017-12-04
Added support for isPressed/isReleased on SDL and Flash

pperidont:


ncannasse:
21 seems hardcoded :)  buttons.length maybe ?
Also, check if it works well with my latest changes (JS support)

PR #317, opened 2017-12-11
Add some (return) types

markknol:
It's easier when figuring out what kind of type is required, also avoids issues like [this](https://github.com/HeapsIO/heaps/issues/302).


ncannasse:
I'm curious : does not completion shows these already?

markknol:
Well I sometimes lots of `unknown<0>`, so its mostly for my own readability.

PR #324, opened 2017-12-28
Allocation optimizations

pperidont:
- Removed allocations in h2d.RenderContext.pushTarget()
- Re-use `h3d.Buffer` in SpriteBatch (disposed only when the buffer is too small)
- Fixed `h3d.Buffer` creation in TileGroup (when empty)
- Fixed `Null<Int>` alloc for optional args in Engine.setRenderZone()

### DIFF
--- a/h2d/SpriteBatch.hx
+++ b/h2d/SpriteBatch.hx
@@ -95,6 +95,7 @@ class SpriteBatch extends Drawable {
 	var last : BatchElement;
 	var tmpBuf : hxd.FloatBuffer;
 	var buffer : h3d.Buffer;
+	var bufferVertices : Int;
 
 	public function new(t,?parent) {
 		super(parent);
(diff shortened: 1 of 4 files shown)


h2d/SpriteBatch.hx
```diff
@@ -290,7 +289,17 @@ class SpriteBatch extends Drawable {
 			}
 			e = e.next;
 		}
-		buffer = h3d.Buffer.ofSubFloats(tmpBuf, 8, Std.int(pos/8), [Dynamic, Quads, RawFormat]);
+		bufferVertices = pos>>3;
+		if( buffer != null && !buffer.isDisposed() ) {
+			if( buffer.vertices >= bufferVertices ){
```
ncannasse: Could we also dispose in case we have halved the number of vertices as well ?

PR #330, opened 2018-01-12
Add support for more keys

nachoverdon:
This adds key codes for the following keys:

Comma ( , ) as COMMA
Fordward slash ( / ) as FSLASH
Semicolon ( ; ) as SEMICOLON
Opening bracket ( [ ) as LBRACKET. L is from Left.
Closing bracket ( ] ) as RBRACKET. R is from Right.

I also sorted some key codes up and down.

But there's a problem, though. There are 2 keys that have conflicts with other keys. I commented those in the code, so they are not implemented yet.

QUOTE ( ' ) key code is the same as RIGHT, 39.
DOT ( . ) key code is the same as DELETE, 46.

### DIFF
--- a/hxd/Key.hx
+++ b/hxd/Key.hx
@@ -16,20 +16,10 @@ class Key {
 	public static inline var HOME		= 36;
 	public static inline var LEFT		= 37;
 	public static inline var UP			= 38;
-	// QUOTE (') keyCode is the same as RIGHT
-	//public static inline var QUOTE	= 39;
 	public static inline var RIGHT		= 39;
 	public static inline var DOWN		= 40;
-	public static inline var COMMA		= 44;
 	public static inline var INSERT		= 45;
-	// DOT (.) keyCode is the same as DELETE
-	//public static inline var DOT		= 46;
 	public static inline var DELETE		= 46;
-	public static inline var FSLASH		= 47;
-	public static inline var SEMICOLON	= 59;
-	public static inline var LBRACKET	= 91;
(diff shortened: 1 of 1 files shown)


ncannasse:
I don't think these has been tested for support on all platforms supported by Heaps (JS, HL/SDL, HL/DirectX, Flash). Also, these keys might have different places depending on the keyboard so I expect a few issues with them.

Will keep open a little more if you wish to contribute full support for it.

nachoverdon:
I stopped as soon as I realized **that there are conflicts with other keys**. This only happens on HL with SDL. This is a problem that I don't know how to tackle, so I leave that to you. Flash and HL with DX use the same key codes, and JS uses most of them as well, without any conflicts.

I tested this with a couple of keyboard and layouts, Spanish and US layouts, both Qwerty, and yes, they are different. Once again, I don't know how to add support for all the possible layouts. **The key codes of this pull request are supposed to support the Qwerty US keyboard layout for now**.

I'm going to update the pull request with the rest of the keys to have basic support... But again, HL/SDL has a problem with conflicting key codes that needs to be fixed in order to add some of the keys. I commented those in the code. I didn't add the Windows keys because I guess that would be different on Linux/MAC...? And the PRINT_SCREEN key is only registred on HL/SDL.

ncannasse:
Several things are necessary to fix for this to work:
- keycodes should not be platform-specific
- all supported platforms and keyboard layouts should be supported.
closing for now

PR #357, opened 2018-03-08
JS: Window touch events prevented

lbergman:
I have a game with a HTML UI layer, and is trying out heaps as renderer, however interactions with the UI layer is blocked on mobile.  Seems like preventDefault is added to all touch actions on the window object, instead of only on the canvas.

EDIT: After looking closer, I cannot really understand why the element should ever be set to window rather than canvas. I can try to refine the PR if I better understood a scenario where one would want to hijack the window events.

### DIFF
--- a/hxd/Stage.js.hx
+++ b/hxd/Stage.js.hx
@@ -26,12 +26,11 @@ class Stage {
 	function new( ?canvas : js.html.CanvasElement ) : Void {
 		eventTargets = new List();
 		resizeEvents = new List();
-
-		element = canvas == null ? js.Browser.window : canvas;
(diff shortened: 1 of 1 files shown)


ncannasse:
This is intended: the idea is to get the keyboard events on a game without requiring the user to click on the canvas first. In order to restrict events to the canvas only, you can simply do the following before creating your hxd.App:

```
new hxd.Stage(myCanvasElement).setCurrent();
```

lbergman:
Thanks! Just had to add @:access(hxd.Stage) as well since constructor is private, but now it works.

PR #374, opened 2018-04-27
Add Polygon.projectPoint and Polygon.distanceSq

trethaller:


ncannasse:
distanceSq : maybe we should add a boolean in/out (if null tests both)

PR #377, opened 2018-05-08
Implemented Bresenham's line algorithm in hxd.BitmapData

TheTrueCaligari:
A quick test code (not sure I'm handling the `h2d.Bitmap` /  `h2d.Tile` generaration correctly)

~~~haxe
class  Main extends hxd.App
{
    var canvas : hxd.BitmapData;
    var bmp : h2d.Bitmap;
    var timer : Float;

    override function init() {
        canvas = new hxd.BitmapData(400, 400);
        bmp    = new h2d.Bitmap(null, s2d);
        bmp.x  = s2d.width/2;
        bmp.y  = s2d.height/2;
    }

    override function update(delta:Float) {
        timer += delta / this.get_wantedFPS();

        canvas.lock();
        canvas.clear(0xFF000000);

        canvas.line(50,50,350,50, 0xFFFF0000);
        canvas.line(350,350,350,50, 0xFFFF0000);
        canvas.line(350,350,50,350, 0xFFFF0000);
        canvas.line(50,350,50,50, 0xFFFF0000);
        
        // PI radians in 30 seconds
        var angle = timer * Math.PI / 30.0 - Math.PI/2;
        
        canvas.line( 200, 200, 
            Math.round(200 + 150*Math.cos(angle) ), Math.round(200+150*Math.sin(angle)),
                0xFFFFFF00);
        canvas.unlock();

        bmp.tile = h2d.Tile.fromBitmap(canvas).center();
    }

    static function main() {
        new Main();
    }
}
~~~


ncannasse:
That seems correct. Maybe adding out of bounds checks to start directly in bitmap and stop when we are exiting it for instance if we draw a line from (-1000,-1000) to (1000,1300)

ncannasse:
There's a trace, and I'm not sure the Math.round are correct wrt keeping the actual line shape.
for instance if you have a almost-but-not vertical line, you still want to be able to see the place bresenham switch pixel y as if you had a bigger canvas/bitmap

TheTrueCaligari:
Oups sorry, those were just tests meant for me...
I did not think the commit would appear in the pull request (its my first time using github like this)

I see what you mean about the Math.round.
So it would be preferable to add the checks in the loops (maybe looping without drawing first until we get in view, and stopping early as you said)

TheTrueCaligari:
I've added some checks to avoid drawing outside of the bitmap.

After a few tests, I used the code described [here](https://stackoverflow.com/a/40902741) which is similar to my last tests but avoided some code redundancies.

When calculating the instersections with the Bitmap border, if the coordinates are too big (arount 2^16 and more...) there are risks of overflowing because of a product, so I used `Int64` as a workaround.
Not sure if it was the best solution.

ncannasse:
Please use Float instead of haxe.Int64 as it's quite emulated on many platforms so produces a lot of extra checks / is slower.

ncannasse:
Up ? :)

TheTrueCaligari:
Sorry... Since the end of August (and the end of my holidays) I had no time to correct this.
But I've noted your remark and will try to get on it as soon as possible.

TheTrueCaligari:
My bad, I forgot to check that.
It should be ok now :)

ncannasse:
Thanks!

PR #383, opened 2018-05-31
Avoid redundant problemMatcher list in tasks.json

Gama11:
I think it's good practice to omit these, even if VSCode generates them automatically. If done so, the defaults from vshaxe's task generation will be used, which means it's more future proof if the list of problem matchers there ever changes. It also makes for a more consise task definition.

### DIFF
--- a/samples/templates/.vscode/tasks.json
+++ b/samples/templates/.vscode/tasks.json
@@ -1,18 +1,10 @@
 {
-	// See https://go.microsoft.com/fwlink/?LinkId=733558
-	// for the documentation about the tasks.json format
 	"version": "2.0.0",
 	"tasks": [
 		{
 			"type": "hxml",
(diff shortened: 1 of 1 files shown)


samples/templates/.vscode/tasks.json
```diff
@@ -1,18 +1,10 @@
...
 			"label": "Build",
 			"file": "::name::_hl.hxml",
-			"problemMatcher": [
-				"$haxe-absolute",
-				"$haxe",
-				"$haxe-error",
-				"$haxe-trace"
-			],
 			"presentation": {
 				"reveal": "never"
```
Gama11: Alternatively, you might want to include a `settings.json` with a `haxe.taskPresentation` setting. That would then also work for other tasks if you have more than one `.hxml` file for instance.

PR #386, opened 2018-06-09
Adding support for Hiero .fnt format

starburst997:
Support for [Hiero](https://libgdx.badlogicgames.com/nightlies/runnables/runnable-hiero.jar) from [libGDX](https://github.com/libgdx/libgdx)

### DIFF
--- a/hxd/res/BitmapFont.hx
+++ b/hxd/res/BitmapFont.hx
@@ -64,10 +64,10 @@ class BitmapFont extends Resource {
 		case 0x6F666E69:
 			// support for Hiero format
 			// https://github.com/libgdx/libgdx
-			entry.getBytes().toString().split('\n').iter(line -> {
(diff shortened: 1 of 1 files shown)


ncannasse:
How complete/stable it is ? (before merging)

ncannasse:
Referencing #445 as it's related.

Yanrishatum:
Looks like we both implemented same format support. Implementation details slightly differ, but this one have multiple issues:
* Does not account for format containing multiple spaces between records.
* Does not account for string field containing spaces while being enclosed in `"..."` (e.g. `face="El Messiri"` would be parsed as `"El` by this version)
* Does not parse `baseLine`
* Does not parse kerning pairs
* Parses `xadvance`, but for some reason uses `width - 1` in place where it should be used.
* For some reason ignores space character.

starburst997:
Oh yeah, it was a basic implementation worked for my use case but not complete

PR #390, opened 2018-06-21
JS text rendering glitch

lbergman:
Using ttf fonts, many of the fonts we use get rendered weirdly with neighbouring chars bleeding over. 
Seems like height sometimes is not picked up properly, and some width margin is needed to ensure glyphs are intact.
We are running with this modification on 9 games with 8 languages and 5 different fonts, and have not seen any issues on any devices after applying this fix.

### DIFF
--- a/hxd/res/FontBuilder.hx
+++ b/hxd/res/FontBuilder.hx
@@ -226,18 +226,18 @@ class FontBuilder {
 		return font;
 	}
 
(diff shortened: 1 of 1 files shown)


ncannasse:
@lbergman seems like getFontHeight function uses spaces instead of tabs, could you update?

lbergman:
Yes, sorry about those...fixed.

ncannasse:
Merged, thanks !

PR #394, opened 2018-07-04
added drawEllipse to h2d.Graphics

josuigoa:


ncannasse:
Maybe would be more complete with an additional "rotationAngle" parameter ?

josuigoa:
oops, I had forgotten this! You mean a "rotationAngle" to rotate the entire ellipse, right?

ncannasse:
Yes 😉

PR #403, opened 2018-08-01
[WIP] Implement binary FBX support

Yanrishatum:
Expands FBX parser capabilities to work with binary FBX files.  
There are limitation on properties, which are:
* Int64 properties stored as Floats.
* Raw binary data property read as String.

Both are due to lack of corresponding FbxProp enums.

Information on parsing was based on this article: https://code.blender.org/2013/08/fbx-binary-file-format-specification/

### DIFF
--- a/hxd/fmt/fbx/BaseLibrary.hx
+++ b/hxd/fmt/fbx/BaseLibrary.hx
@@ -1,4 +1,5 @@
 package hxd.fmt.fbx;
+import haxe.io.Bytes;
 using hxd.fmt.fbx.Data;
 import h3d.col.Point;
 
@@ -148,7 +149,7 @@ class BaseLibrary {
 		defaultModelMatrixes = new Map();
 	}
 
-	public function loadTextFile( data : String ) {
+	public function loadTextFile( data : Bytes ) {
 		load(Parser.parse(data));
 	}
 
--- a/hxd/fmt/fbx/Parser.hx
+++ b/hxd/fmt/fbx/Parser.hx
@@ -160,15 +160,15 @@ class Parser {
 		
 		var props : Array<FbxProp> = new Array();
 		var childs : Array<FbxNode> = new Array();
-		for (i in 0...numProperties) {
+		for ( i in 0...numProperties ) {
 			props.push(readBinaryProperty());
 		}
 		
-		if (pos < nextRecord) {
+		if ( pos < nextRecord ) {
 			do {
 				childs.push(parseBinaryNode());
 			}
-			while (pos + 13 < nextRecord); // There is null-record of 13 bytes afterwards for some reason.
+			while ( pos + 13 < nextRecord ); // There is null-record of 13 bytes afterwards for some reason.
 		}
 		
 		return { name: name, props: props, childs: childs };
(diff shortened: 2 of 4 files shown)


ncannasse:
How much WIP is this ? 
I would recommend transforming this Int64 into Haxe Float instead of Int, will get more digits representation. That's what we're doing for > 9 digits integer in text parser.

ncannasse:
Also, did you find any part that has raw binary ? you can add PBinary( haxe.io.Bytes ) if you want.

ncannasse:
I'll merge for now, we will test it at Shiro, don't hesitate to submit patches if you get some fixes

jcward:
Oh, very exciting! :+1: 

Hmm... I tried a few exports from Blender (a simple set of primitives, and this [demo buggy model](https://download.blender.org/demo/test/AtvBuggy.zip) (this one had an overflow I had to ignore)) -- and while the FBX->HMD conversion worked, nothing shows up on the screen (just hacking them into samples like world and interactive.)

Yanrishatum:
@jcward I managed to get the model rendered with exception of messed up culling (it seem to be modelled with expectation for backfaces not being culled), no material colors and there are two sets of tires for some reason. Looks like you need to manually apply the modifiers before exporting, which is strange, because there is a checkbox that should apply them automatically during export.
img
Furthermore, I could not get that model to render in Autodesk FBX Converter, plus I get same rendering results as shows in screenshot after converting binary FBX to ASCII, which means it's not a binary parser issue, but general one :)
(Binary parser in general just reads file data, and interpretation of it still the same as with ASCII)

ncannasse:
@Yanrishatum if you had to disable culling it's maybe because the model was exported with wrong axises. Heaps assumes FBXs are exported in Right-handed - Z up axis, and convert them to Left Handed Z up. I haven't look how the exported axis are encoded into FBX, we might be able to support multiple export options if handled correctly.

Yanrishatum:
I exported it with Z up, and -X forward. If I'd export it with wrong up axis, it'd be oriented the wrong way.

ncannasse:
We just found out you need Z up , Y forward or else X and Y axis are flipped. It's normal to have X axis being flipped between Blender and Heaps because Heaps is using Left Handed coordinates whereas Blender is using Right Handed.

jcward:
Hmm, I tried disabling culling (I don't know how -- like this?)

image

and exporting with Z Up, Y Forward -- still nothing on my screen (world sample --> JS, trying to replace tree with buggy, or cubes, or anything out of Blender):

image

ncannasse:
@jcward I also noticed that blender models are scaled 100x , I have fixed that on GIT version, but maybe your model is too small or too big to be seen in the world.

kevinresol:
I have a FBX binary (v7.5) model on hand and it doesn't parse properly (got a root node with empty name, empty property list and empty children list)

[model.fbx.zip](https://github.com/HeapsIO/heaps/files/2263274/model.fbx.zip)

Yanrishatum:
@kevinresol I'm looking at the FBX, and there's a lot of padding. So far I've tested only on version 7400. Primary thing is that FBXHeaderExtension have 0 offset. (Next chunk offset, paramCount, paramSize, nameLen(byte), name, nodeList. Looks like I need to change how node list parsed and accept that chunk offset may be 0, because node right after root node have 0.

Yanrishatum:
Okay, seems like 7500 changed format quite a bit, and I'll have to rewrite parser quite significantly to introduce support of it. Will be in another PR most likely.

PR #413, opened 2018-08-29
Outline filter

lbergman:
Basic outline filter.

Needed a filter to use for simple outlines, that can be thick but still sharp.
Seems like existing outline shader will not work as a 2D shader, so I created an 'Outline2D' shader (not sure about the name since it might also be usable as a 3D shader, let me know if you have a better idea) and filter and pass similar to Blur.

I feel a bit out of my depth here, so let me know if there is anything that can be done in a better way. Especially the pass.Outline.apply function is basically a copy paste job from Blur where i removed the parts related to normals and cubes to keep it simple for now. I'll be happy to add them back, but haven't got time right now to set up test examples for those to ensure they work as expected, so left it out for now.


lbergman:
Seems to be broken in WebGL 1...working on a fix.

lbergman:
Works now on ios and in ie, (although IE refused to start at all with latest heaps, but have a separate PR on the way for that)

ncannasse:
Hi Leo, can't you use a Glow ? It should produce a nice outline when well setup.

lbergman:
It doesn't seem to be possible to get it to produce a smooth thick outline with no alpha using Glow.

If I use large radius with high alpha, I do get a thick outline, but it doesn't follow text tightly enough for our designer, and does leave a glow effect which is not desired.

Outline filter also seems to perform much better in our usage.

Pixi has a glow which is very easy to configure to work as an outline with separate strength and distance params, but still has a separate outline filter which performs much better for basically same effect (which is where I got the fragment code from):
http://pixijs.io/pixi-filters/tools/demo/

Here is an extreme example, using Outline:
outline_rainbow
 
And the closest I can acheive by using Glow:
glow_rainbow

However, of course we have no need to push this into heaps, just thought others might find it useful as well.

ncannasse:
The visuals tell it all, merged ;)

PR #420, opened 2018-09-09
Save tile groups for tilesets

Beeblerox:
It's easier to get tile groups properties from CdbLevel than to parse them manually afterwards


ncannasse:
Should "LevelGroup" have an Array of tiles or a single tile of the whole size ? I think the later would be more accurate

ncannasse:
Thx!

PR #425, opened 2018-09-23
Add custom cursor support to JS target

Yanrishatum:
Method of operation:
Uses `url(data: ...)` to set the cursor image. To produce base64 of the bitmap, native `canvas.toDataURL` used instead of Haxe built-in features due to `haxe.zip.Compress` not being implemented on JS.
Animated cursors can be supported by switching the frames, but I'm not sure where the hook for update should be set up. (Same method may also be used to animate cursor on HL target?)

### DIFF
--- a/hxd/Cursor.hx
+++ b/hxd/Cursor.hx
@@ -24,6 +24,8 @@ class CustomCursor {
 	static var UID = 0;
 	var name : String;
 	var alloc : flash.ui.MouseCursorData;
+	#elseif js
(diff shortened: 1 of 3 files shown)


Yanrishatum:
I'm not sure adding it via Haxe MainLoop would be a good idea, because I need the `hxd.Timer.deltaT` for timer as well as it should be performed and the end of Heaps primary tick. MainLoop will make it inconsistent, as there is no guarantee that it will be called after the Heaps tick.

PR #429, opened 2018-09-28
Add support for animated cursor to JS, HLDX and HLSDL targets.

Yanrishatum:
Should've used separate branch for that...
Adds `haxe.MainLoop` function that updates cursor. Since it utilizes same update code between JS and HL, it was implemented in custom cursor class, and not in `System.hx`.
Known issue: On JS target cursor sometimes blink (on chrome at least), and I'm not sure what causes it.


ncannasse:
Seems like there was a commit problem wrt line endings, the whole files seems modified, can you update?

Yanrishatum:
Fixed. Note to self: Do not use github integrated merge tool, it messes up files.

Issue #431, opened 2018-09-29
Audio on JS target is extremely bad


Yanrishatum:
Right now JS target built-in audio support produces just atrocious sound quality. I've included reproduction code, and will explain issues seen in just that simple sample:
* ~Audio is clearly distorted is louder than it supposed to be.~ - fixed in #432 
* If you try to play it multiple times, it will start to pre/post cut sound. Keep in mind it's not an issue of how many sounds plays simultaneously, as during LD42 I had this issue in main menu, where no other sounds were used (music played via native HTMLAudio), and no more than 5 could be played at once.
* There is noticeable pause between click and sound actually playing. - UPD: This is a chromium bug, and 
 I already working on solution to reduce lag from ~210ms to at least 100ms. Chrome for some reason uses empty buffer on script processor node before inquiring sample data.
* Bonus not in the sample: For music (long tracks, 2+ minutes long) I couldn't even get it load properly, it usually never started, and at occasions when it did, sound quality was atrocious. 

Reproduction code:
[hljsaudio.zip](https://github.com/HeapsIO/heaps/files/2430094/hljsaudio.zip)

lbergman:
When I played around with heaps audio on JS I found it unusable since script nodes have a memory leak on many browsers, so that is something to watch out for as well.

For example:
https://bugs.chromium.org/p/chromium/issues/detail?id=379753

On some level I really like the approach of emulating OpenAL in WebAudio...great way to make audio consistent across targets...but with the current state of browser support for script nodes it is impossible to use heaps audio on JS in my experience, at least if you have a decent amount of triggers over time, in which case app will run out of memory quickly.

Yanrishatum:
@lbergman I poked around a bit and implemented pool of created nodes, so they are reused instead of being created when possible. I believe it should partially fix the leak issue (at least I no longer get 400+mb usage after running sounds ~2000 sounds, and only temporary raise in usage with it returning to normal after GC). Downside is... Well, it now holds the pool of maximum sounds you used at once :)

PR #432, opened 2018-09-29
Fix audio quality issues on JS target

Yanrishatum:
Poor audio quality on JS was caused by resampling of F32 data to I16, resulting in extreme precision loss.
It's a partial fix of #431, but response time and early cut issues still present.


ncannasse:
Response time might be because there's a delay in HTML API for loading and decoding sound files. They can be avoided by preloading the sounds but at the cost of more cpu / memory usage at app startup.

PR #435, opened 2018-10-02
Add workaround to chromium memory leak for js audio.

Yanrishatum:
Adds pooling of created ScriptProcessorNodes, to avoid memory leaks. As in chromium: https://bugs.chromium.org/p/chromium/issues/detail?id=379753
Reported by @lbergman here: https://github.com/HeapsIO/heaps/issues/431#issuecomment-425915172

### DIFF
--- a/hxd/snd/NativeChannel.hx
+++ b/hxd/snd/NativeChannel.hx
@@ -105,6 +105,10 @@ class NativeChannel {
 		}
 		return ctx;
 	}
+	// Pool implemented to workaround memory leak in chromium browsers:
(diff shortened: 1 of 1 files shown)


lbergman:
Nice!
I guess the drawback will be that only one instance of a sound can be played at the time now?
IMO better than the leak, but worth noting the change in behaviour.

Yanrishatum:
No, it's not the case. Everything works the same way. Change is made to workaround the memory leak with keeping a pool of created nodes, trying to obtain them from here first, and only then fallback to creating a new one. After the sound completed playback, node returns back to pool. It does not limit the sound system usage in any way.

PR #437, opened 2018-10-04
Rework JS NativeChannel

### LINKED ISSUE #431: Audio on JS target is extremely bad

Yanrishatum:
Okay, this is quite large internal change. Primary reason to such drastic change - poor browser support of ScriptProcessorNode. (Memory leaks, asynchronous method of operation, latency, etc. Also it's deprecated in favor of AudioWorklets).
Current approach pretty much uses same method as ScriptProcessorNode, but instead relies on `AudioBuffer` and `AudioBufferSource` for playback, which are a lot more stable.
Implementation notes:
* `time` variable is mandatory to sync back-to-back buffer playback, otherwise they will have an audible gaps due to `ended` event being asynchronous, and `AudioContext.currentTime` being slightly ahead in time.
* It always fills both buffers at start (bufferSamples*2), I'm not sure how that'll work out for extremely short sounds that fit in just one buffer.
* Channel buffer size increased to 8192 to fix sound distortion in Firefox. It does not introduce more latency, as it is no longer affects it due to data being buffered ahead of time.

This resolves next issues:
* Enormous sound latency on Chrome.
* Large sound distortion when multiple sounds played at once. No idea why ScriptProcessor had that issue, but BufferSource doesn't.

Unresolved issues:
- [x] [Firefox] Some sounds still have glitching. See this [image](https://i.imgur.com/MvlxMKg.png) for reference. To the left is firefox (central part have "glitch"), to the right is the correct waveform, produced by Chrome. Note: Distortion appears on 4096 buffer size, disappears on 8192 size and sound straight out broken on 2048.
- [ ] [Firefox] Looping of sounds have noticeably more distortion if the loop isn't perfect. -- Wontfix
- [x] Not sure if I have to pool created AudioBuffers. Avoids excessive allocs, but is really necessary?
- [x] When playing multiple sounds, it does not distort them it was before, but it now cuts some sounds early, seems like previous sound playback stops current sound playback for some reason. 
- [ ] When tab is not active, sound playback will stop. On chrome it just cuts, but on firefox for some reason it sometimes plays some buffers occasionally.

It should fix #431 latency and multiple sounds issue.

### DIFF
--- a/hxd/snd/NativeChannel.hx
+++ b/hxd/snd/NativeChannel.hx
@@ -108,13 +108,13 @@ class NativeChannel {
 	// Avoid excessive buffer allocation when playing many sounds.
 	// bufferSamples is constant and never change at runtime, so it's safe to use general pool.
 	static var pool : Array<js.html.audio.AudioBuffer> = new Array();
+	static var bufferPool : Array<haxe.io.Float32Array> = new Array();
 	
 	var front : js.html.audio.AudioBuffer;
 	var back : js.html.audio.AudioBuffer;
 	var current : js.html.audio.AudioBufferSourceNode;
 	var queued : js.html.audio.AudioBufferSourceNode;
-	var time : Float;
-	// TODO: Pool tmpBuffer as well?
+	var time : Float; // Mandatory for proper buffer sync, otherwise produces gaps in playback due to innacurate timings.
 	var tmpBuffer : haxe.io.Float32Array;
 	#elseif lime_openal
 	var channel : ALChannel;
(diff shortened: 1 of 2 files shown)


Yanrishatum:
@ncannasse I'd say it's safe to merge after discussing the part that I still use pooling, but this time to store unused `AudioBuffer`s. Is it worth it to cache them? If yes, should I also cache the created temporary buffers (`tmpBuffer`) as well that are used to populate buffer data? I just not really fond of the idea that in sound-heavy environment allocation of all those buffers can be pretty intense, but it's more of performance/memory footprint choice issue.

Yanrishatum:
@ncannasse Just wanna ping that it's pretty much finished and ready for merge. If I'll do further improvements on that matter, they won't be soon.

ncannasse:
Thanks! Merging now.

PR #441, opened 2018-10-08
Add support for non-relative 2D particles

Yanrishatum:
Feature requested in [this topic](https://community.heaps.io/t/make-particles-position-not-relative-to-its-parent/39).
Adds support to make particles non-relative.
Implementation details:
I apply position difference between initial global position of particle and new global position of particles objects. Calculations are done once per particle init and once per frame per group (multiple particle groups would invoke `Object.localToGlobal` multiple times and initialization of object would call it `nparts` time per group).
There also an issue with some particles "twitching" while in this mode, most likely due to conversion local->global->local->global and I'm not sure if I'll be able to solve this issue.

### DIFF
--- a/h2d/Particles.hx
+++ b/h2d/Particles.hx
@@ -54,7 +54,7 @@ private class ParticleShader extends hxsl.Shader {
 
 @:access(h2d.ParticleGroup)
 private class Particle extends h2d.SpriteBatch.BatchElement {
-
+	
 	var group : ParticleGroup;
 	public var vx : Float;
 	public var vy : Float;
@@ -63,6 +63,9 @@ private class Particle extends h2d.SpriteBatch.BatchElement {
 	public var maxLife : Float;
 	public var life : Float;
 	public var delay : Float;
+	
+	public var originX : Float;
+	public var originY : Float;
 
 	public function new(group) {
 		super(null);
@@ -86,8 +89,8 @@ private class Particle extends h2d.SpriteBatch.BatchElement {
 		vx += group.gravity * et * group.sinGravityAngle;
 		vy += group.gravity * et * group.cosGravityAngle;
 
-		x += vx * et;
-		y += vy * et;
+		x = (originX - group.compensate.x) + vx * et;
+		y = (originY - group.compensate.y) + vy * et;
 		life += et;
 
 		if( group.rotAuto )
@@ -112,7 +115,7 @@ private class Particle extends h2d.SpriteBatch.BatchElement {
 
 		if( group.animationRepeat > 0 )
 			this.t = group.tiles[Std.int(t * group.tiles.length * group.animationRepeat) % group.tiles.length];
-
+		
 		if( t > 1 ) {
 			if( group.emitLoop ) {
 				@:privateAccess group.init(this);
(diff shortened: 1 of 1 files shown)


ncannasse:
That seems more complex than it should be. 

What about instead setting the particle position as relative to parent when emit, then upon draw overwrite the Particles matrix before write and restore it after write?

Yanrishatum:
It would require reworking how SpriteBatch rendering works. Right now particle code have no way of affecting it's matrix directly, as it's done asynchronously. (It first updates the BatchElement and only then does flush which in turn applies the matrix) Hence the complicated implementation - SpriteBatch internally does not support non-relative elements.

ncannasse:
@Yanrishatum if you look at Particles class, it uses drawWith(batch,this) which means it's the Particles matrix that gets used, and it could be swapped at this time before calling drawWith, the restored back afterwards

Yanrishatum:
@ncannasse Ah, I see. Then the implementation is indeed extremely straightforward. Thanks!

ncannasse:
@Yanrishatum about about scale/rotation ? :)

ncannasse:
I think we should reset to identity, it can be kept per group. Also, we might want to apply the full matrix to the particle direction at emit time when !isRelative

Yanrishatum:
Done and done.

h2d/Particles.hx
```diff
@@ -351,6 +356,21 @@ class ParticleGroup {
...
 		p.maxLife = life;
+
+		if ( !isRelative ) {
+			// Less this.parts access
+			var parts = this.parts;
+			// calcAbsPos() was already called, because during both rebuild() and Particle.update()
+			// called during sync() call which calls this function if required before any of this happens.
+			//parts.syncPos();
+
+			p.x += p.x * parts.matA + p.y * parts.matC + parts.absX;
```
ncannasse: I don't think this should be += but simply =

h2d/Particles.hx
```diff
@@ -351,6 +356,21 @@ class ParticleGroup {
...
+			var parts = this.parts;
+			// calcAbsPos() was already called, because during both rebuild() and Particle.update()
+			// called during sync() call which calls this function if required before any of this happens.
+			//parts.syncPos();
+
+			p.x = p.x * parts.matA + p.y * parts.matC + parts.absX;
+			p.y = p.x * parts.matB + p.y * parts.matD + parts.absY;
+			p.scaleX = Math.sqrt((parts.matA * parts.matA) + (parts.matC * parts.matC)) * size;
+			p.scaleY = Math.sqrt((parts.matB * parts.matB) + (parts.matD * parts.matD)) * size;
+			p.rotation += Math.acos(parts.matA / p.scaleX);
```
ncannasse: I'm still a bit unsure about that Math.acos here. I think an Math.atan2(matB / scaleY, matA / scaleX) would be better but I haven't tested it.

Yanrishatum: Yes, you're right (it does back-and-forth movement which I couldn't notice at first), also noticed another bug with px/py calcualation. Due to `p.x` being assigned first, it's new value affects `p.y`, while it supposed to use old one. Plus I overlooked the need to rotate movement velocity as well.

ncannasse:
Thanks!

Issue #367, opened 2018-04-16
Font glyphs vertical alignment


ncannasse:
NG Title fonts are slightly offset in Y compared to their top-left corner, maybe because we should align the glyphs based on the lower letters (gj etc.). This is something that needs study/fix.

Yanrishatum:
Is it still relevant? Since Heaps now supports better BMF format that provides font baseline among other things. There's not much point in improving clearly inferior format, especially since manual already extensively covers font file generation and advises against using Divo.

ncannasse:
I think it is relevant. Your font tool still has the same baseline guess wrong I think ? Correct me if I'm wrong.

Yanrishatum:
No, it does not.
For fonts in BMF format, baseline is already provided by the generator, and code that does guessing kicks in only if `baseLine == 0`, which happens only on Divo format or fonts that for whatever reason have baseline configured incorrectly.
See `common.base` in bmf spec: https://www.angelcode.com/products/bmfont/doc/file_format.html
Same goes for fontgen - it uses baseline value provided by the font file and stores it in the bmf:
https://github.com/Yanrishatum/fontgen/blob/master/native/msdfgen_lib.cpp#L92

ncannasse:
Thanks for the precision, closing.

PR #352, opened 2018-02-27
Added simple binary format for Atlases and Fonts

pperidont:
- Added support for binary format in hxd.res.BitmapFont and hxd.res.Atlas
- Added converters
- Automatically converts fonts and atlases in hxd.fmt.pak.Build
- Fixed kerning inversion in FontBuilder/devo parser

PR #445, opened 2018-10-15
Improve BitmapFont processing

Yanrishatum:
This PR gives some love to the font support. As littera was working with [BMF format](http://www.angelcode.com/products/bmfont/), I figured introducing more extensive support of it.

* Added support for binary and text BMF.
* Fixed parsing of XML to not rely on generator faithfullness to specs. While Littera does not generate `<?xml ...` header, BMFont does, hence the change. New detection checks if `<font>` node contains `<info>` node which present only in BMF, but not in Divo.
* Littera XML: Parse baseLine. Divo does not provide `baseLine`, but is present in BMF. As a result, `baseLine` guessing required only for Divo. 
* Text sample: Showcase of format support.

Not sure, but I think #367 refers to baseline calculation being imprecise. Due to this changes, all supported formats except Divo should have this issue resolved.

Additional notes: BMF also have support for multiple texture atlases (pages). I did not made support to it, but ideally it should be able to support it. (And it would simplify the text sample, because it would use less PNG images ;) )

UPD: Closes #386 as more comprehensive implementation.


ncannasse:
While you're at it, you might want to look at #352 : the idea would be to treat fonts descriptors the way we treat fbx files, by preprocessing them and creating a single fast-to-load binary file. Tell me if instead you just want me to merge this PR.

Yanrishatum:
Implemented BFNT writer. It uses different structure, orienting on fast parsing directly into `h2d.Font`. But due to this, I had to replicate variables and some functions `h2d.Font`, `h2d.Font.FontChar`, `h2d.Font.Kerning` and `h2d.Tile` due to Convert being called in macro context, and type parsing of `h2d.Font` leads to resolving of `@:build` in `h3d.mat.Pass` which is disallowed in macro context.

ncannasse:
f158249374808c7e1c6c403aab32a98745ec6bae should help with classes redefinitions : you should now be able to use h2d.Font from within macros

ncannasse:
Good for merge. Now next logical step would be to contribute the corresponding documentation here:
https://github.com/HeapsIO/heaps-doc/blob/master/Text.md

PR #446, opened 2018-10-16
Basic video support on JS target.

### LINKED ISSUE #444: Video support for HTML
It seems that video is only currently supported on Hashlink.
Would be great to get support for this for html5 also.

Yanrishatum:
Provides basic implementation of h2d.Video on JS target.
Tested on WebGL2 compatible browsers and abomination that shouldn't have seen light of day - Edge (WebGL1).
What's lacking:
* ~Video is set on loop, it never actually invokes `onEnd`~
* Video should be provided as an url and cannot be bundled as part of embedded or pak resources.
* ~There is no error handling (e.g. if video loading fails)~
* Adds out-of-spec function to `GlDriver` on JS target to allow rendering of the video element onto texture.
* No audio output provided.

Overall this implementation feels a bit hacky for me personally, so I'd be glad to receive some feedback. Especially if some of the provided above points should be fixed before merge. 

Fix #444

### DIFF
--- a/h2d/Video.hx
+++ b/h2d/Video.hx
@@ -149,7 +149,11 @@ class Video extends Drawable {
 	#if js
 	
 	function errorHandler(e : js.html.Event) {
+		#if (haxe_ver >= 4)
 		onError(v.error.code + ": " + v.error.message);
+		#else 
+		onError(Std.string(v.error.code));
+		#end
 	}
 	
 	function endHandler(e : js.html.Event) {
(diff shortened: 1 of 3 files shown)


ncannasse:
Thanks !
Small nitpicking : I would prefer uploadTextureVideoElement to be private and accessed with @:privateAccess from within h2d.Video

ncannasse:
For your other remarks, I think that's already a good start people can contribute upon. Is there some way to have the `onEnd` working?

PR #448, opened 2018-10-16
Implement wireframe support

### LINKED ISSUE #407: Wireframe
It's been requested several times the ability to render as wireframe. While Stage3D does not support it other platforms should be able to do so.

Yanrishatum:
* Added: `driver.hasFeature(Wireframe)`
* Wireframe mode can be set per-material pass with `Pass.wireframe` flag.
* Samples: Added checkbox to toggle wirefram on top box for Base3D sample.
Implementation concerns:
* DirectX and GL wireframes produce different results: 
OpenGL:
image
DirectX:
image
Not sure which one should be considered required (I suppose GL)
* On GL `drawMode` never resets, I'm not entirely familiar with rendering pipeline and concerned if it may corrupt some draw calls that do not use `selectMaterial`?

Resolves #407


ncannasse:
Can you confirm you can't get backface culling in wireframe mode in DirectX ? Seems strange.
(try agains https://github.com/HeapsIO/heaps/commit/640d8976a7ae60591004cf83883ac6fb9eda8ae0 which prevents bit clashing)

ncannasse:
(Also, drawMode handling seems ok to me)

Yanrishatum:
Set culling to forced None when fireframe mode enabled on DX, and changed drawMode for wireframe on GL. They still render slightly differently, but I don't think I'll be able to fix that by simple draw mode mangling. :) (for WebGL there's no other way to render wireframe, but for OpenGL - there is [`glPolygonMode`](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glPolygonMode.xhtml) which can be utilized for proper wireframe rendering on SDL)

ncannasse:
Seems good, thanks!

PR #459, opened 2018-10-24
Add h2d Camera

### LINKED ISSUE #458: Introduce 2D camera
I've noticed a lot of people ask about how to do camera in s2d. Usually people just refer to Beeblerox's [heaps samples](https://github.com/Beeblerox/Simplest-Heaps-Examples/tree/master/13_heaps_camera_scroll). Since it's such a frequent thing people want - it just begs a question on why there is no camera controller for s2d out of the box?
Hence, I propose adding `h2d.Camera` class, but there's much to discuss:

* How exactly camera would affect the scene? Two obvious ways of implementation (both have issues)
  * Beeblerox method: Make camera an Object and add objects it should display as a child. 
Issue:: Additional layer between object and scene, no multiple cameras support.
  * Make camera manipulate Scene position. 
Issues: Most likely requires a lot of work on Scene/Layers to account for it's transformation, no multiple cameras support. 
* Should there be support for multiple cameras? (Picture-in-picture, etc)
* Should camera support post-process effects?
* Should camera allow for rotation and scale?

Yanrishatum:
Implementation details:
* Camera acts as a tree node that can be only added to `h2d.Scene` or `h2d.MultiCamera`.
* Uses additional shader matrix to apply after calculation of absolute position.
* Added `h2d.ParallaxObject` that have `scrollX` and `scrollY` values for camera. See docs for details.
* Extremely underdeveloped support for multiple cameras (ver/hor ratio, MultiCamera).
* Basic Camera2D sample.

Todo:
- [x] Improve Camera2D sample.
  - [x] Make it look adequate instead of eye-burning white noise.
  - [x] Showcase `h2d.ParallaxObject`.
  - [x] Showcase `h2d.Interactive` with conjunction to camera movement.
- [x] Add `follow` variable to camera. (Allow following specific objects, no interpolation).
- [x] `h2d.Interactive` following camera working correctly with the mouse.
- [x] Postprocessing on Camera? (hxd.Object filters)
- [x] Camera rotation/scale.
- [x] Add parallaxing object.
- [x] Implement MultiCamera - Basic version implemented, but not proper one. Probably will do with separate PR.
- [x] `anchor` that sets which part (corners/center) of the camera should be used as an anchor when screen size is changed and camera adapts it's center position. Also used as anchoring point for followed object.
- [x] `ratio` variables to scale display-area of camera. Not a mask. Used for anchoring when scene is resized.


ncannasse:
I'm not sure that's the good approach. As we discussed in #458 I think Camera should be an `h2d.Object` in which you add things.

Regarding multiple viewports for same content, I think this should be handled by having a MultiCamera class that have several views and does the rendering / sync X times

Yanrishatum:
On the contrary, I object against making camera as tree object. Imagine a scene with roughly 10k objects. If camera to move each frame, it would cause non-stop recalculation of all objects positions causing big strain on CPU. I am against such implementation, since it is too much unnecessary actions on slight camera state change.

ncannasse:
Well that could be optimized by having cameraX/Y affect directly the final rendering matrix in RenderContext, all objects would not be marked as posChanged

AustinEast:
If this method of providing a Camera is added, I would consider taking a page from Haxeflixel with how it's objects are affected by Camera Scrolling. 

Changing the `followCamera` boolean to a `Point` called `scrollFactor`, and using that point in your `absolutePosition` calculations would allow you to specify how much each object is affected by the scrolling. 

For example, an object with a `scrollFactor` of `(1,1)` would scroll with the camera, an object with a `scrollFactor` of `(0,0)` wouldn't scroll at all, and an object with a `scrollFactor` of `(2,1)` will scroll twice as fast horizontally, but will scroll with the camera vertically. Doing things this way will essentially give you Parallax Scrolling as well :)

Hopefully that made sense!

Yanrishatum:
@ncannasse 
E.g. you want to make camera a tree object, but instead of having just Camera in Scene it would affect it indirectly while rendering it's content? That seem rather hacky, as it will have to keep it's own matrix separate from h2d.Object which always should be identity. And in the end where exactly it should apply the matrix? `baseShader.absoluteMatrix`? That would be basically same as it is right now, but camera matrix applied on CPU for each object instead of GPU and there will be extra object in the rendering tree.

@AustinEast
That's what I meant by `scrollX` / `scrollY`.

Yanrishatum:
I've reimplemented camera to be part of object tree, but it can be added only to Scene. It still uses shader matrix approach. Also implemented `scrollX/Y` instead of `followCamera`. Basically absolute minimum that's left is to add support for Interactive and fix rotation, plus improve sample a bit. But CameraController and MultiCamera would be good to add too ;)

ncannasse:
Sorry for the delay in reviewing. I'm fine with the changes but without the camScrollX/Y on h2d.Object. The main reason is that I want to keep that Camera system optional and have as low impact as possible on runtime when it's not used.

Another reason is that this makes a difference between the logical coordinates (getBounds() , localToGlobal etc.) and the real display ones. I taking that into account at every level would be too costly/complex.

One alternative would be to provide a `CameraParallax extends h2d.Object` which would change current camera position relatively to some factors before rendering its children.

Yanrishatum:
And then people will start massively ask how to do parallaxing. :) But okay, I'll remove it. 
Yet I should disagree on too costly/complex. It's just multiplication per object on a tree and pretty cheap cached calculation within `calcAbsPos`.

ncannasse:
It also takes 32 additional bytes of memory per Object, and my second remark still applies ;) Not trying to argue, just making my thoughts more understandable (hopefully)

JoeCreates:
With multiple cameras, we would also need a way to set some objects in the world to display on one camera but not another, e.g. an archer player has a line showing their arrow's projected path only visible on their screen. Would the proposed design support this?

Yanrishatum:
@JoeCreates 
No, at least not in the first iteration of camera support. I actually not even sure I'll make the MultiCamera at all, as it requires substantial amount of work for Interactive to work with multiple cameras. Most likely there will be a few more iterations of the camera system to introduce multiple cameras, Interactive support for them and camera-exclusive objects. (I have an idea on how to do it, but I need to think on it first)

Yanrishatum:
Bump, as it's pretty much finished PR.

samples/Camera2D.hx
```diff
@@ -0,0 +1,88 @@
...
+			if ( v ) {
+				camera.follow = followObject;
+			}
+			else {
+				camera.follow = null;
+				camera.x = camX;
+				camera.y = camY;
+			}
+		});
+		addSlider("Anchor X", function() { return camera.anchorX; }, function(v) { camera.anchorX = v; }, 0, 1);
```
markknol: Very minor thing, but in Haxe you can also just do:
```haxe
addSlider("Anchor X", function() return camera.anchorX, function(v) camera.anchorX = v, 0, 1);
````

markknol: No not important, just a comment.

ncannasse:
Sorry for the delay :)

I took time to review the whole PR. While I understand the need for a Camera system, I think the current implementation brings some issues that will cause other problems.

For instance getBounds(), filters out-of-screen cuts, etc. rely on absolute positions being always in screen space, not in camera space.  I'm not much happy also having the Interactive depends on having a camera or not, etc.

I also think that recalculating a few dozens of objects positions when the camera moves is not a big problem compared to actually drawing them, or testing if they are out of screen, etc.

I would thus prefer the patch to be changed by having the camera inherit its transformation to sub objects. If you still want to have a very special optimization for camera translation, you can still find a way to recursively translate all non-moving sub objects matrixes if only an offset has been applied to the camera.

Yanrishatum:
Sigh. Okay. I'm done. I won't try to persuade you that this is a bad decision. I'll just delete all the stuff I worked on and do as you say. Also less features, because I don't want to implement stuff like following object or parallax objects with THAT kind of camera. I'm done arguing.

ncannasse:
@Yanrishatum sorry, please don't take it badly, I usually prefer when we reach consensus but I fail to find a way to accommodate both our points of views together. Thanks anyway for the final PR.

Softwaerewolf:
Why would a camera be a parent container for the objects it's rendering? I've literally never seen any engine do this and for good reason, if the camera is the container then the game state is strongly-coupled to the camera and you also have to duplicate the entire object tree to have a second camera (eg. for splitscreen) which is wildly inefficient.

ratkingsminion:
May I also comment that the current 2D camera system is the worst I ever experienced? :) Cameras look at things; they're not the parent of everything in the world ...

shinaka:
I'd also like to throw my vote in for the camera being the parent to all objects making no practical sense at all. Regardless of any attempt to justify the decision, there isn't an engine out there that behaves this way (arguably for good reason), so this is never going to meet user expectations. 

Honestly, I had been considering using Heaps for a small project to try it out, but the state of this kind of stuff (and extremely patchy documentation) is turning me off of the idea. :/

ncannasse:
The current camera was a first attempt. Further redesign is being discussed
in #622

PR #462, opened 2018-10-30
Add some docs for Flow object

Beeblerox:


h2d/Flow.hx
```diff
@@ -54,6 +57,15 @@ class FlowProperties {
...
 }
 
+/**
+ * Container object which behaves like flexbox container in CSS.
+ * Could be usefull when you want to place objects without worrying about setting their coordinates manually 
+ * (for example, in for automatic layout of UI elements).
+ * This object will take care about it, so none of the added object will overlap others. Each object will be placed after another,
+ * and when `width` or `height` (depends on `isVertical` property value) becomes bigger than `maxWidth` or `maxHeight` 
+ * then new line or new column will be started to place next objects.
+ * You'll just need to call `rewlow()` method and all objects will be positioned for you.
```
markknol: `rewlow`.

(Also try to avoid "you" in documentation.)

Beeblerox: ok, will try to rework it, but my English isn't good at all

h2d/Flow.hx
```diff
@@ -54,6 +57,15 @@ class FlowProperties {
...
 }
 
+/**
+ * Container object which behaves like flexbox container in CSS.
+ * Could be usefull when you want to place objects without worrying about setting their coordinates manually 
+ * (for example, for automatic layout of UI elements).
+ * This object will take care about it, so none of the added objects will overlap others. Each object will be placed after another,
+ * and when `width` or `height` (depends on `isVertical` property value) becomes bigger than `maxWidth` or `maxHeight` 
+ * then new line or new column will be started to place next objects.
+ * Don't forget to call `reflow()` method after adding/removing objects from container and all objects will be positioned automatically.
```
ncannasse: It is not necessary to call reflow() manually.

h2d/Flow.hx
```diff
@@ -54,6 +57,15 @@ class FlowProperties {
 
 }
 
+/**
+ * Container object which behaves like flexbox container in CSS.
+ * Could be usefull when you want to place objects without worrying about setting their coordinates manually 
+ * (for example, for automatic layout of UI elements).
+ * This object will take care about it, so none of the added objects will overlap others. Each object will be placed after another,
+ * and when `width` or `height` (depends on `isVertical` property value) becomes bigger than `maxWidth` or `maxHeight` 
+ * then new line or new column will be started to place next objects.
```
ncannasse: only if isMultiline=true, so maybe move this part to isMultiline doc

h2d/Flow.hx
```diff
@@ -78,11 +90,26 @@ class Flow extends Object {
 	**/
 	public var verticalAlign(default,set) : Null<FlowAlign>;
 
+	/**
+	 * Min width. Flow's width won't be less than this value, even when the total width of added objects is less than this value.
```
ncannasse: No necessary to repeat "Min width" here (same for others)

h2d/Flow.hx
```diff
@@ -845,4 +873,4 @@ class Flow extends Object {
 	public dynamic function onAfterReflow() {
 	}
```
ncannasse: I think it would be necessary to fully document the class before merge, as the style/vocabulary needs to be consistent.

ncannasse:
Thank you for the contribution! However the documentation is not yet precise enough for me to merge it as-it. I'll make sure to find some time to edit it before merging.

PR #463, opened 2018-11-03
Add `shape` support to `h2d.Interactive`

Yanrishatum:
Basically allows for refined hit-test on mouse for `h2d.Interactive` with usage of `h2d.col`. With a sample.

What I don't like: It's still limited to width/height parameters. But removing them would make it a breaking change. Solutions are:
* If shape is present, ignore width/height and use shape bounds.
* Remove those outright and just have the shape as input, same as `h3d.Interactive`. + introduce Rectangle shape.
* Or just merge it as is, because it just works already, but is a pain to use, because you have to sync shape position and width/height manually, otherwise some parts of shape won't be checked for collision.

Also not sure if it need the `preciseShape` or not.

### DIFF
--- a/h2d/Scene.hx
+++ b/h2d/Scene.hx
@@ -49,6 +49,7 @@ class Scene extends Layers implements h3d.IDrawable implements hxd.SceneEvents.I
 	var window : hxd.Window;
 	@:allow(h2d.Interactive)
 	var events : hxd.SceneEvents;
+	var shapePoint : h2d.col.Point;
 
 	/**
 		Create a new scene. A default 2D scene is already available in `hxd.App.s2d`
@@ -61,6 +62,7 @@ class Scene extends Layers implements h3d.IDrawable implements hxd.SceneEvents.I
 		height = e.height;
 		interactive = new Array();
 		eventListeners = new Array();
+		shapePoint = new h2d.col.Point();
 		window = hxd.Window.getInstance();
 		posChanged = true;
 	}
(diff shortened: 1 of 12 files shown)


h2d/Scene.hx
```diff
@@ -280,6 +289,11 @@ class Scene extends Layers implements h3d.IDrawable implements hxd.SceneEvents.I
 			}
 			if( !visible ) continue;
 
+			if (i.shape != null) {
+				pt.set((kx / max) * i.width + i.shapeX, (ky / max) * i.height + i.shapeY);
```
ncannasse: maybe set pt after having the event.relX and event.relY are set, makes less code duplication.

h2d/col/BitmapCollider.hx
```diff
@@ -0,0 +1,64 @@
...
+
+	/**
+		If true, will collide if any channel is above cutoff. Otherwise will collide only if all channels above their cutoff values. (default : true)
+	**/
+	public var collideOnAny : Bool;
+
+	/**
+		Create new BitmapCollider with specified bitmap, channel cutoff values and check mode.
+	**/
+	public function new(bitmap: hxd.BitmapData, alphaCutoff:Int = 127, redCutoff:Int = 255, greenCutoff = 255, blueCutoff = 255, collideOnAny = true) {
```
ncannasse: hxd.BitmapData has some platform specific quirkness, maybe you should use pure-cpu hxd.Pixels instead ?

h2d/col/Collider.hx
```diff
@@ -0,0 +1,7 @@
+package h2d.col;
+
+interface Collider /* extends hxd.impl.Serializable.StructSerializable */ {
```
ncannasse: why the commented extends?

Yanrishatum:
Question about relX/relY conflict wasn't answered as well as width/height issue. ;)

ncannasse:
I would like to keep a width/height for the Interactive, but yes I understand when it's a shape it does make so much sense.

What we could do would be to ignore width/height testing when Interactive has a shape. We shouldn't try to optimize by getting shape bounds and instead directly check for contains(). Some smart shapes will be able to check for bounds first. That also means that when there's a shape, we need to recalculate relX/Y by inverting the Interactive 2D matrix transform. This can maybe be cached.

PR #468, opened 2018-11-07
Fix runtime JS parsing crash on invalid codepoint on Raw binary property

### LINKED ISSUE #456: Blender 3D Models, FBX 6.1.0
Hello! I'm new to Heaps and i was testing some models made in Blender. Blender only exports to FBX 6.1.0, while Heaps only supports FBX 7.1.0. Is there a way to load FBX 6.1.0 models into Heaps? And if not, can you create some kind of support for it?

Yanrishatum:
Bug was caused due to JS specifics and `readString` tried to read it into cohesive valid String. I disabled handling of `R` property type (Raw binary data) - it isn't used anywhere where it maters. Right now it resolves to empty `PString`. If someone will need to parse those properties for some reason - better file an issue. (It will require adding new enum - PBinary, and subsequent handling of it in all places where this enum is checked)
Fixes #456

### DIFF
--- a/hxd/fmt/fbx/Data.hx
+++ b/hxd/fmt/fbx/Data.hx
@@ -7,6 +7,7 @@ enum FbxProp {
 	PIdent( i : String );
 	PInts( v : Array<Int> );
 	PFloats( v : Array<Float> );
+	PBinary( v : haxe.io.Bytes );
 }
 
 typedef FbxNode = {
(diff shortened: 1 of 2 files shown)


ncannasse:
Let's instead add a proper PBinary, even if unused.

Yanrishatum:
Done. Looks like I overestimated the usage of this enum, as all `switch` statements check for specific enum types and always have `default:`.

PR #469, opened 2018-11-07
Improve h2d.Interactive collision with shape.

Yanrishatum:
* Ignore width/height if `shape` is set.
* Add invDet cache to h2d.Interactive
* Simplify `eventToLocal`
* Disable `isEllipse` for shapes.
* Some documentation.
* Visibility check was relocated before collision check.

### DIFF
--- a/h2d/Scene.hx
+++ b/h2d/Scene.hx
@@ -252,31 +252,6 @@ class Scene extends Layers implements h3d.IDrawable implements hxd.SceneEvents.I
 			var i = interactive[idx];
 			if( i == null ) break;
 
-			var dx = rx - i.absX;
-			var dy = ry - i.absY;
-
-			var w1 = i.width * i.matA;
-			var h1 = i.width * i.matC;
-			var ky = h1 * dx + w1 * dy;
-
-			// up line
-			if( ky < 0 )
-				continue;
-
-			var w2 = i.height * i.matB;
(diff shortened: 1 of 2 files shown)


h2d/Scene.hx
```diff
@@ -289,13 +264,48 @@ class Scene extends Layers implements h3d.IDrawable implements hxd.SceneEvents.I
 			}
 			if( !visible ) continue;
```
ncannasse: I would like to keep width/height testing before visibility testing, in case there are many Interactive within deep hierarchies.

Yanrishatum: And for case of shape interactive? Should it check for visibility before or after shape collision check?

Yanrishatum:
Oh, and another thing. Maybe `getBounds` should actually get shape bounds if it's present instead of `width` / `height` ?

h2d/Scene.hx
ncannasse: I guess we should be uniform here and check after in all cases

ncannasse:
Regarding getBounds it's a bit tricky.
setting backgroundColor will not work with shape (would require to be able to draw that shape)
and for "forSize", it's only when the interactive is inside a Flow, and usually you have an additional graphics that goes with it.

ncannasse:
Thanks!

PR #475, opened 2018-11-14
HtmlText: improved "htmlToText" function

Klug76:
HtmlText: improved "htmlToText" function:
1. Added option "condenseWhite" like in flash::TextField.
2. Repeated erroneous replacement of html entities is removed (they are already replaced in Xml.parse)

### DIFF
--- a/h2d/HtmlText.hx
+++ b/h2d/HtmlText.hx
@@ -4,6 +4,8 @@ import h2d.Text;
 
 class HtmlText extends Text {
 
+	public var condenseWhite(default,set) : Bool = true;
+
 	var elements : Array<Object> = [];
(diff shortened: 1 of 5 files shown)


ncannasse:
Regarding HTML entities, could you confirm it's the case on both JS and HL ?

Klug76:
Yes.
Let's assign the variable `var singleText = "Test: &amp;lt;";` in `build/htmlText/HtmlText.hx`. Expected result is `Test: &lt;`: [proof](https://www.w3schools.com/code/tryit.asp?filename=FXCFC16CROT2). Result in the master branch: `Test: <`: 
image

ncannasse:
Thanks!

PR #476, opened 2018-11-17
Robust interface to add Converts.

Yanrishatum:
Right now, there's basically no robust way to add Convert to the FileSystem to be executed at compile-time. This PR aims to introduce robust and easy to understand way to do so.
I used same approach as with `hxd.res.Config.extensions` - as in, provide the Convert class.

### DIFF
--- a/hxd/fs/Convert.hx
+++ b/hxd/fs/Convert.hx
@@ -2,6 +2,12 @@ package hxd.fs;
 
 class Convert {
 
+	/**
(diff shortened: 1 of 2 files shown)


ncannasse:
I think this require a bit more thinking.  We are using LocalFileSystem when doing hxd.Res.initEmbed and it runs in macro mode, so we cannot call macros from it.

Yanrishatum:
Should be fixed now.

ncannasse:
Thanks

PR #479, opened 2018-11-19
Add support for format-tiled library.

### LINKED ISSUE #478: Tiled format support out of date
Hello, I'm chekcing out some parts of the lib and am noticing the Tiled support is a bit out of date.
I guess that because CDB is starting to become something pretty interesting and gathers leverage here, I suppose Tiled won't be supported as well as before.

So the helper class lags behind the current version of the file structure (notably lacks stuff like group layers, tile collisions, etc...)

Should it be kept, even though it lags behind features, should it be deprecated, should it be moved to a small helper library?

(I'm also wondering how to properly make the class evolve. I have my own implementation based on the .json format that does a bit more than the library's class (like generating a collision mesh), but it's probably a bit too specialized?)

Have a nice day.

Yanrishatum:
Current `hxd.res.TiledMap` is quite basic, and there is already pretty extensive library that provides good support for Tiled. This PR adds support for [format-tiled](https://github.com/Yanrishatum/haxe-format-tiled) library with preserving legacy parser if no library present. 
Heaps-specific implementation:
* Expects up-to-date library (2.0.0+)
* Automatic resolving of TSX references.
* Automatic loading of tileset textures (both regular texture.
* Optionally can apply tile type templates if provided with.

Logical second iteration would be a class that deals with layer rendering, but it requires a lot of work, and I'm not sure I would ever take on this task. :) 

Resolves #478

### DIFF
--- a/hxd/res/TiledMap.hx
+++ b/hxd/res/TiledMap.hx
@@ -1,4 +1,98 @@
 package hxd.res;
+
+#if (format_tiled >= "2.0.0")
+
+import format.tmx.Data;
+import format.tmx.Reader;
+using format.tmx.Tools;
+
+typedef TiledMapData = {
+	var tmx : TmxMap;
+	/** Optional list of loaded tilesets when loading map with `loadTilesets = true`. **/
+	var tilesets : Array<TiledMapTileset>;
+}
+
+typedef TiledMapTileset = {
+	var tileset : TmxTileset;
+	/**
+		List of all tiles in the tileset.
+		Note that they are not guaranteed to share the same texture, if tileset is an image set.
+	**/
+	var tiles : Array<h2d.Tile>;
+}
+
+class TiledMap extends Resource {
+
+	var reader : Reader;
+
+	/**
+		Parses TMX file and optionally resolves TSX references and loads tileset images. objectTypes can be provided to add their properties to objects.
+	**/
+	public function toMap(resolveTsx = true, loadTilesets = true, ?objectTypes:Map<String, TmxObjectTypeTemplate>) : TiledMapData {
+		
+		reader = new Reader();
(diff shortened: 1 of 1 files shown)


Yanrishatum:
Moving it to `hxd.fmt` is not an option, because otherwise I'll have to maintain two versions of the library, and I won't do that. And I'm not fond of moving it to general `format` due to irregularity of library updates, and even if I'd introduce it to main library, it would take a while to get in haxelib. Additionally, then people will start to get errors about lack of `tmx` package just because their library is older. Lastly, maintaining it would be less convenient. 
I see it more like optional dependency, akin to hxbit or castledb (and I, for example, use neither)

ncannasse:
If the Heaps support depends quite strongly on format-tiled changes, then maybe you should add it there as well ? We would then have only a stub in Heaps redirecting to your implementation. Not sure if it's the best, I don't have a strong opinion about how to organize things here as they are multiple dependent layers.

Yanrishatum:
It doesn't depend on the changes that much, I just targeted the 2.0.0+ version, since it got multiple breaking changes compared to 1.X.X. (Mainly to improve reader usability, and custom parameters support)
I prefer to avoid doing breaking changes, and PR should be compatible with all 2.X.X versions that will follow. I don't see the point of making stubs either, because `#if` specifically checks for compatible version. (Or I misunderstood what you meant by doing a stub)
To be perfectly honest, I don't really care if that PR going to be accepted or not, I just noticed very poor and outdated support for Tiled and added optional capability to use library specialized on processing Tiled files as well as add Heaps-specific processing of image data, described above. (If rejected, I'll just relocate it to my "too complex for Heaps" library :) )

Yanrishatum:
Closing this in favor of my Heeps extension library.

Issue #482, opened 2018-11-23
HL touch handling


josuigoa:
I'm facing some issues with the recent changes of HashLink/SDL touches.

When I touch the screen, SDL fires two events: MouseDown and TouchDown. And from those events Heaps creates two EPush events that don't make any difference.

This should be handled by the user or would Heaps return a single event?

ncannasse:
It should return a single event

josuigoa:
A fix that comes to my mind is to ignore mouse events `if (Sys.systemName() == 'Android')` in the haxe side, but it doesn't seem very elegant to me. Maybe there is a more convenient way to do it in SDL? What do you think?

ncannasse:
in `hxd/Window.flash.hx`, we did use `hxd.System.getValue(IsMobile)`  but it's not yet implemented correctly in `System.hl.hx`, so I guess you could submit a PR for both fixes.

ncannasse:
Close by #485

PR #485, opened 2018-11-27
touch events in HL

josuigoa:
related to #482


ncannasse:
I think this might require some caching in order to avoid calling Sys.systemName for each event processed.

ncannasse:
Thanks !

PR #488, opened 2018-11-29
Keep ParticleGroup getFields to not break on dce

lbergman:


lbergman:
Closing since fix should not be needed in Haxe 4-rc.1

PR #490, opened 2018-11-30
WIP: Fix delayed position update for non relative particles

lbergman:
Fixes https://github.com/HeapsIO/heaps/issues/489

### DIFF
--- a/h2d/Particles.hx
+++ b/h2d/Particles.hx
@@ -80,6 +80,9 @@ private class Particle extends h2d.SpriteBatch.BatchElement {
 				visible = true;
 			else {
 				visible = false;
+				if(!group.isRelative) {
+					group.init(this);
+				}
 				return true;
 			}
 		}


Yanrishatum:
This would call `group.init` each frame when particle is not visible effectively changing delay each frame (I'm a bit surprised it even shows particles at this point). 
In order to properly fix it:
* Add `ParticleGroup.reset`:
```haxe
function reset( p : Particle ) {
	p.delay = hxd.Math.random() * life * (1 - emitSync) + emitDelay;
	if ( isRelative || p.delay <= 0 )
		init(p);
}
```
* `ParticleGroup.rebuild` should call `reset(p)` instead of `init(p)`.
* When particle delay reach 0:
```haxe
if ( !group.isRelative )
	group.init(this);
```
* When particle lifetime ends, it should call `group.reset(this)` instead of `group.init(this)`.
* `ParticleGroup.init` should not set `delay` anymore.

With this approach main particle initialization happens only once - as before for relative particles and on delay expiry in non-relative mode.

ncannasse:
Please update the PR :)

lbergman:
Sorry for the delay. We headed in different direction and don't need this fix right now, and I haven't had time to spare to look at it. If you like, please go ahead and close this PR, and I can open a new one when I have made a proper fix.

Issue #486, opened 2018-11-28
Lime library dependency in LimeStage.Keyboard


mepsoid:
When I'm trying to use constants for key codes described in `hxd.impl.LimeStage.Keyboard` class, compiler warns `LimeStage.hx:209: characters 38-53 : Type not found : lime.ui.KeyCode`.

Heaps version 1.5.0, Haxe version 4.0.0

mepsoid:
`hxd.Key` is incomplete, for example it doesn't contain 187 or 189 key codes. If `Keyboard` class not used for heaps project building it can be excluded fron library at least to not duplicate constants.

Yanrishatum:
Then create an issue (or even PR) about lacking key codes, maybe? ;)
To explain why these classes are present: Lime backend is not officially supported because nobody maintains it. What is present although - already implemented partial support to Lime which have no guarantees to work. Hence, `hxd.impl.LimeStage.Keyboard` _is_ used, but only if you use Lime backend. Same as classes that tied for HLDX, HLSDL and JS targets.

PR #331, opened 2018-01-17
Update Key.hx

markknol:
Close #325

PR #492, opened 2018-11-30
Add more keycodes to `hxd.Key`

### LINKED ISSUE #325: Add support for more keys
I didn't see keys like comma, period, semicolon, etc. on the keys list that I think should be added.

Yanrishatum:
Third time the charm, amiright? ;)
Adds following key code constants:
* Equals <kbd>=</kbd>
* Minus <kbd>-</kbd>
* Tilde <kbd>~</kbd> / <kbd>`</kbd>
* Left and right brackets <kbd>[</kbd> <kbd>]</kbd>
* Punctuation <kbd>;</kbd> <kbd>'</kbd> <kbd>,</kbd> <kbd>.</kbd>
* Slashes <kbd>\\</kbd> <kbd>/</kbd> + `\` located next to left shift on some keyboards (`INTL_BACKSLASH`)
* Windows and Context Menu keys
* <kbd>CapsLock</kbd>, <kbd>ScrollLock</kbd>, <kbd>NumLock</kbd> and <kbd>Pause/Break</kbd> keys. 
* Extended function keys support from <kbd>F1</kbd>-<kbd>F12</kbd> to <kbd>F1</kbd>-<kbd>F24</kbd>.

SDL specifics:
I mapped all keys except `INTL_BACKSLASH`. Because hlsdl uses `e.key.keysym.sym` instead of `e.key.keysym.scancode`, and I cannot differentiate between them. Scacodes are 0x31 and 0x64, but keycode is always 0x5C. 
In fact, I'd say hlsdl require some works to pass both scancode and keycode. Because if I'll switch my layout to Russian, keycode will report a letter of that layout (For example: ` = ё, ' = э, etc) while scancode remains persistent. This is required to allow both text input and keybinds not reliant on active keyboard layout.
The F13+ keys are mapped, but I could not make SDL actually react to them, it never receives the key event whenever I press F13/F24. Not sure why.
Because PrintScreen was only reported by SDL, by not other targets I decided to not include it.

I tested HLDX, HLSDL, JS and Flash targets to ensure it works consistently on all of them. But only on regular QWERTY keyboard. I'd appreciate if someone will test it on AZERTY/QWERTZ keyboard.

Related: #330, #331, #486.
Closes #325


ncannasse:
> * Equals =
> * Minus -
> * Tilde ~ / `
> * Left and right brackets [ ]
> * Punctuation ; ' , .
> * Slashes \ / + `\` located next to left shift on some keyboards (`INTL_BACKSLASH`)

I'm not sure all of them makes sense when it comes to supporting several keyboard layouts.
For instance on AZERTY, both `;` and `.` are on the same key, and `~` can only be input with an Alt+2 combination, etc.

So I'm not sure how to deal with these. We could maybe have a specific prefix, such as QWERTY_TILDE, and then this way we can add other i18n keyboards specific keys support as well.

Keep in mind that key codes should NOT be used for text input. We have a specific event for that that should be able to deal with specific type of inputs that doesn't consists in key/scan codes (soft keyboard, IME, etc.)

Yanrishatum:
I want to set some specifics straight. `hxd.Key` supposed to represent physical key locations, which are persistent from keyboard to keyboard, right? Current implementation lacks a lot of keys, and while their actual charcode differ from layout to layout - same keycode should always have same position on keyboard. And using QWERTY key names is the most logical, since it's basically standard keyboard layout for the world, while QWERTZ/AZERTY are language-specific.
I want to stress main point - `hxd.Key` should be independent of current active keyboard layout. Be it QWERTY, AZERTY or QWERTZ or whatever. When I press Q in english, Q in russian or A in french layout - it always should report `Key.Q`. If it depends on layout - it will lead to stupid issues when player switch layout and suddenly half of hotkeys just do not work. 
Due to that, I'm marking this as WIP, because I'll need to dig more on how to implement this to be independent of layout. So far I'll need:
- [ ] HLSDL: (Library-side) Given that, I will double down on the SDL using `scancode` instead of `sym`, because `sym` depends on layout, while `scancode` isn't.
- [ ] HLDX: (Library-side) Find how to pass proper scancode.
- [ ] Flash: (Heaps-side) Find how to pass proper scancode.
- [ ] JS: (Heaps-side) Process `event.code` string to extract scancode.
Additionally it worth checking out on detecting active layout and return according key name.

nachoverdon:
Thanks :)

ncannasse:
> "hxd.Key supposed to represent physical key locations, which are persistent from keyboard to keyboard, right?"

When they use the same layout yes, but for instance there is no TILDE key on AZERTY. Instead we have a `²` key at the same place which has code 222 in HLDX and 178 in HLSDL. And at the same physical location on a QWERTY keyboard you get different code... 

Some other keys don't really "translate" well because the actual location might differ from keyboard to keyboard even with same original layout (think about all the laptops moving some keys around)

> I want to stress main point - hxd.Key should be independent of current active keyboard layout. Be it QWERTY, AZERTY or QWERTZ or whatever. When I press Q in english, Q in russian or A in french layout - it always should report Key.Q

I'm not sure I agree with this. For instance if you implement Ctrl+Z as hxd.Key.isDown, you don't want it to become Ctrl+W in AZERTY. Handling keyboard layouts is not easy, I would prefer if Heaps does not try to be too smart by doing things that can solve some problems but create as many.

So I would prefer if we keep the current behavior where key names match the current keyboard layout. As for non-standard keys, we can add constants and names with layout prefix so developers remember it's a problem to deal with at some point.

Yanrishatum:
Done.

ncannasse:
Thanks!

Issue #494, opened 2018-12-04
Initializing JS window with a canvas should prevent events from being globals to the window (not working)


deepnight:
Main.hx:
```
	static function main() {
		#if js
		var canvas : js.html.CanvasElement = cast js.Browser.document.getElementById("webgl");
		var w = new hxd.Window(canvas, false);
		w.setCurrent();
		#end
		new Main();
	}
```

If the browser window isn't big enough (ie. it has scrollbars), if you press UP/DOWN while the canvas has focus (ie. clicked on it), the page scrolls while the client also receives these inputs.

Expected:
the browser shouldn't receive keyboard events while the client is focused, like any web component.

deepnight:
I've sent a fix :)

PR #496, opened 2018-12-04
Block keyboard events if the canvas is "focusable" (ie. element==canvas)

deepnight:
Block keyboard events if the canvas is "focusable" (ie. element==canvas) so they don't spread to the document if the canvas is focused.

See issue #494 
[Issue](https://github.com/HeapsIO/heaps/issues/494)

### DIFF
--- a/hxd/Window.js.hx
+++ b/hxd/Window.js.hx
@@ -54,6 +54,10 @@ class Window {
 		if( element == canvas ) {
 			canvas.setAttribute("tabindex","1"); // allow focus
 			canvas.style.outline = 'none';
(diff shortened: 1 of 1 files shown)


deepnight:
For some reasons, things like h2d.Console no longer receive keyDown event. I don't see why a stopPropagation here should do such a thing, while the onKeyDown() method is still called in Window.js.hx .

Yann-P:
Your goal is to prevent scrolling right?
For me, it is sufficient to add `if(focused) e.preventDefault();` as first line of `onKeyDown`.

PR #504, opened 2018-12-16
Improve Interactive (both h2d/h3d)

### LINKED ISSUE #502: H2d Interactive OnRelease bug
I have an app I'm making with heaps and I have multiple bitmaps with unique functions attached to them that are triggered with an OnRelease.

Picture like a 2x2 array of bitmaps where each number is a unique bitmap

[Func1] [Func2] [Func3] [Func4]
[Func5] [Func6] [Func7] [Func8]

I find if I push down on one bitmap and then drag and release, both the first initial bitmap's event is triggered and the one where the function was released in but none of the ones in between.

Ie, If I click on Func1, then drag through Func2 and release in Func7, both OnRelease events are triggered for Func1 and Func7

Yanrishatum:
* Added documentation to `onPush`, `onRelease`, `onReleaseOutside`, `onClick`, `onOver` and `onOut`
* Removed usage of `ERelease` kind during `onRelease` when event is `EReleaseOutside`
* Add `onReleaseOutside` event that fires instead of `onClick` if released outside Interactive hitbox area.

Personal remark: Due to lack of documentation it was unclear how `onRelease` worked, and the fact that it used `ERelease` while in reality event was `EReleaseOutside` didn't help a lot. 
Close #502

### DIFF
--- a/h2d/Interactive.hx
+++ b/h2d/Interactive.hx
@@ -165,9 +165,9 @@ class Interactive extends Drawable implements hxd.SceneEvents.Interactive {
 			mouseDownButton = -1;
 		case EReleaseOutside:
 			if( enableRightButton || e.button == 0 ) {
-				e.kind = ERelease;
 				onRelease(e);
-				e.kind = EReleaseOutside;
+				if ( mouseDownButton == e.button )
+					onReleaseOutside(e);
 			}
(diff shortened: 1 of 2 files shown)


ncannasse:
Thanks!

PR #505, opened 2018-12-18
textBaseline bottom for js target in font builder for more even result across browsers

Sechgulo:
Chrome with their latest [update](https://chromium.googlesource.com/chromium/src/+log/339739f679a9ad4943d77add01513b409f60f416/third_party/blink/renderer/core/html/canvas/text_metrics.cc) adjusted text metric for their baseline value to more correspond to the standard (something only FF has done previously). This however results in varying results across browsers since `top` baseline is calculated differently between browsers. `bottom` however seems to be more even across all browsers (compare this [fiddle](http://jsfiddle.net/j87bY/1/) in FF and Edge/Safari for example). Due to this I suggest using `bottom` as baseline instead.

I've also removed the previously added arbritrary `xMarg` value which was added to prevent artifacts/scrambled chars. Using `bottom` textBaseline seems to prevent previous noted artifacts from occuring.

### DIFF
--- a/hxd/res/FontBuilder.hx
+++ b/hxd/res/FontBuilder.hx
@@ -186,6 +186,7 @@ class FontBuilder {
 				if( size == null ) continue;
 				var w = size.w;
 				var h = size.h;
+				if( h > lineH ) lineH = h;
 				// Whatever the first character is it will always have x and y value 0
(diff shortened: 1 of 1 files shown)


ncannasse:
I'll trust you on this one :)

PR #508, opened 2018-12-22
Add Signed Distance Field fonts support.

Yanrishatum:
- Adds SignedDistanceField basic shader.
- Adds ability for bitmap fonts to use this shader for scalable fonts.

Issues:
- Does not work on flash. I'll be honest on that matter. **I do not care about flash** and do not want to fiddle with trying to get shader working on it and figure out what exactly AGAL do not like. Since 2018, flash must die for real this time around.
- Because of above - Text sample does black screen on flash target.
- I suspect HtmlText won't work adequately with SDF fonts, because of how SDF implemented.


ncannasse:
Thanks for looking into it, I never had the time to implement it :)
A few comments before merge:
- could we replace `sdfAdded` by `sdfShader == null`. This way we only create the shader when necessary
- I am not sure if "buffer" and "gamma" values should be part of the Font itself, or instead parameters for the h2d.Text. How often these needs to be changed in a typical usage? In all cases, they need proper documentation. I think "buffer" does not really need much changes (it's a 0.5 constant in libgdx) so maybe let's keep it only in the shader definition. And gamma should be named "smoothing" the way libgdx does because "gamma" has some other definition (it's usually a pow(2.2) for gamma correction).
- otoh I would make the "channel" parameter an enum (R/G/B/A/MultiChannel) so both normal and multichannel SDF fonts uses the same code paths
- I don't see any reason why HtmlText would not be working
- No problem with flash not working, simply #if !flash the sample part that uses sdf

Yanrishatum:
> could we replace sdfAdded by sdfShader == null. This way we only create the shader when necessary

Then if font gets changed from one SDF font to another SDF font - shader will be added twice, because `addShader` does not check if shader instance was already added. Alternatively, I can just dispose of shader instance, if font is set to regular bitmap font.

> otoh I would make the "channel" parameter an enum (R/G/B/A/MultiChannel) so both normal and multichannel SDF fonts uses the same code paths

I used int, because HXSL (what a surprise) have no documentation, and I have no clue if it's possible to pass enum const. 

> I don't see any reason why HtmlText would not be working

I'm not entirely sure how it operates, and since SDF shader is added to h2d.Text - it may break the rendering of images and such. 

> buffer

I was basing this variable on this article: https://blog.mapbox.com/drawing-text-with-signed-distance-fields-in-mapbox-gl-b0933af6f817
As showcased here, buffer is not really a constant - it dictates the alpha cutoff for the field. While default 0.5 provides glyph form, increasing it, for example can give cheap and pretty accurate outline.
However I can rename it to `alphaCutoff` for cleaner understanding of what it does.

ncannasse:
> Then if font gets changed from one SDF font to another SDF font - shader will be added twice

When setting the font:
- if sdfFont && sdfShader == null : create + add
- if not stdFont && sdfShader != null : remove and set null

>  I have no clue if it's possible to pass enum const.

Yes HXSL does not support enums, it conforms the GLSL specification but with Haxe syntax. You can use an abstract enum or use the enum .getIndex() to pass the index.

> HtmlText

HtmlText creates sub objects for bitmaps and addShader will only affect the current object so I think it's fine atm.

ncannasse:
Should the alpha cutoff and smoothing be part of the font ? I have no idea if they actually describe the font or are some effect on a specific text, in which case we should instead add a single `setSdfStyle( smoothing, cutoff )` method instead.

(ps: you left over sdfAdded field decl)

Yanrishatum:
They describe how font should be rendered. This field should remain more or less static regardless of font scale. `alphaCutoff` represents at which part of sdf lies the edge and `smoothing` dictates how much that edge is smoothed. Smaller value leads to sharper edge, while bigger to more blurred one. Hence it describes how the font itself supposed to be rendered and not some effect on the Text.
Maybe worth exposing override values in `h2d.Text` then?

Hah, totally forgot about removing actual value. I also going to remove flash checks, thanks to #510.

ncannasse:
One last thing I noticed: you are using  output.color  in your shader instead of pixelColor, which means you don't get all other shaders contributions, that's the reason I think you had to pass the color manually.

h3d/shader/SignedDistanceField.hx
```diff
@@ -0,0 +1,42 @@
...
+			var textureSample : Vec4 = textureColor;
+			var distance : Float;
+
+			distance = if (channel == 0) textureSample.r;
+				else if (channel == 1) textureSample.g;
+				else if (channel == 2) textureSample.b;
+				else if (channel == 3) textureSample.a;
+				else median(textureSample.r, textureSample.g, textureSample.b);
+			
+			output.color = vec4(color.rgb, color.a * smoothstep(alphaCutoff - smoothing, alphaCutoff + smoothing, distance));
```
ncannasse: should be pixelColor *= ....

Yanrishatum:
Actually, no. I had to pass it manually, because `pixelColor` from `Base2d` already had `textureColor * color` multiplied, and there is no way to obtain just the tinting data from `Base2d`. At least not that I know of. SDF produces monochromatic image based on texture and then tints it. It does not modify previous pixel color.

Yanrishatum:
Done. Although I don't really understand how exactly it works and sorts out in which order variables calculated. (SDF code executes after Base2d.__init__ where it reads textureColor but before assigns pixelColor? How? Why? Seriously, lack of documentation is big roadblock for using hxsl)

ncannasse:
@Yanrishatum the only "trick" here is that each expression in an "__init__" section is taken as a standalone "program" that will be linked based on order of precedence together with other shaders.

rzer:
Smoothstep is not supported on flash. You can use function:
`function smoothstep(a:Float, b:Float, x:Float):Float{
	var t:Float = saturate((x - a)/(b - a));
	return t*t*(3.0 - (2.0*t));
}`

Yanrishatum:
@rzer Smoothstep is implemented in HXSL AGAL output in #510

PR #509, opened 2018-12-22
Add captureSubPixels

Yanrishatum:
- Adds `Driver.captureSubPixels` function with `bounds:h2d.col.IBounds` argument that allows to capture only region of the texture. Implemented on GL and DX drivers.
- Adds optional bounds argument to `Texture.capturePixels`.

Reasoning:
Sometimes you don't need entire texture pixel data captured, but only specific region of it. This provides robust API to do so.


ncannasse:
Thanks, please make it an optional parameter to Driver.capturePixels instead.

PR #510, opened 2018-12-22
added Smoothstep AGAL operation

Klug76:
This operation is required for new SDF-fonts to work in flash/AIR target.


ncannasse:
Thanks!

PR #514, opened 2018-12-31
Add MP3 support to Hashlink.

Yanrishatum:
Depends on HaxeFoundation/hashlink#214
Allow playback of mp3 files on Hashlink, allowing usage of mp3 as multi-target sound format.

### DIFF
--- a/hxd/res/Config.hx
+++ b/hxd/res/Config.hx
@@ -66,12 +66,12 @@ class Config {
 			if( defined("hl") ) HL else
 			Unknown;
 		switch( pf ) {
-			case HL:
-				#if heaps_enable_hl_mp3
-				ignoredExtensions.set("mp3", true);
-				#end
-			default:
-				ignoredExtensions.set("ogg", true);
+		case HL:
+			#if !heaps_enable_hl_mp3
+			ignoredExtensions.set("mp3", true);
+			#end
+		default:
+			ignoredExtensions.set("ogg", true);
 		}
 		return pf;
 	}
--- a/hxd/res/Config.hx
+++ b/hxd/res/Config.hx
@@ -65,8 +65,14 @@ class Config {
 			if( defined("cpp") ) Cpp else
 			if( defined("hl") ) HL else
 			Unknown;
-		if ( pf != HL )
-			ignoredExtensions.set("ogg", true);
+		switch( pf ) {
+			case HL:
+				#if heaps_enable_hl_mp3
+				ignoredExtensions.set("mp3", true);
+				#end
+			default:
+				ignoredExtensions.set("ogg", true);
+		}
 		return pf;
 	}
 
--- a/hxd/res/Config.hx
+++ b/hxd/res/Config.hx
@@ -65,12 +65,8 @@ class Config {
 			if( defined("cpp") ) Cpp else
 			if( defined("hl") ) HL else
 			Unknown;
-		switch( pf ) {
-		case HL:
-			ignoredExtensions.set("mp3", true);
-		default:
+		if ( pf != HL )
 			ignoredExtensions.set("ogg", true);
-		}
 		return pf;
 	}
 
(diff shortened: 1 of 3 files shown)


ncannasse:
MP3 has several issues in particularly related to looping (see https://github.com/HeapsIO/heaps/blob/master/hxd/snd/Mp3Data.hx#L15)

I would prefer to keep the current workflow ogg-by-default on HL, that makes things more easy, for instance when having a WAV we will encode it to OGG by default, and we are skipping the MP3 files. When you have a project for both HL and Web, we either use WAV for small sounds that gets encoded at compilation or pak-ing, or we have both MP3 and OGG files for musics.

I understand that you might prefer mp3 on hl for your own usage, I think in that case the best if to have it as a separate library `hlmp3`. We can eventually add `#if hlmp3` support in Heaps to support this configuration when the user (you or some others) wants it, and document it as part of Heaps documentation, but I would prefer it to be manual opt-in by adding `-lib hlmp3` to compilation parameters.

Yanrishatum:
I understand that MP3 have issues, but I have to disagree on outright removing its support out of the box. My view is that being able to play different sound formats is a positive and allows people to use whatever workflow they prefer/fine feasible.
Here's an example on why ogg->mp3 workflow is not always optimal.
Ogg is a lossy format, which means conversion of it to mp3 leads to even more quality loss. On top of that, Heaps converter just does not provide the ability to configure at which settings mp3 is encoded. When targeting html5 foremost - it is more logical to output mp3 files directly from the audio software rather than converting it from ogg, but while html5 is an output target, HL is utilized due to it's debugging capabilities. So, manual conversion of mp3->ogg each time file changes? (That's actually exactly what I had to do)

And I have to disagree even more about getting separate library as `hlmp3`, instead of having it in `fmt`. Mp3 is still a format, and it may be useful when not using Heaps and just utilizing HL.
If, for some reason, it's really that preferable to not allow user  to use MP3 while targeting HL, hiding it behind `hl_enable_mp3` flag or just having a warning by default that discourages usage of mp3 as sound files on HL should be sufficient enough.

ncannasse:
We actually never do ogg->mp3 (or mp3->ogg) encoding, we only do wav->mp3 or wav->ogg, or output for music either mp3 or ogg from the software. I agree doing mp3->ogg leads to poor quality, as would ogg->mp3 (here, ogg refers to ogg/vorbis)

We could maybe support a common lossless audio codec such as FLAC and allow conversion from it to both mp3 and ogg, as it would be less of a big file than using WAV as sources.

Please note that for small sounds wav is good enough: it gets compressed to either mp3 or ogg depending on the platform, transparently.

Yanrishatum:
Updated for changes in hashlink/mp3 PR branch.
Additionally added optional define flag `heaps_enable_hl_mp3` that will enable referencing mp3 files from Res. This way, mp3 files can be played back on HL target (which was the purpose of the PR in the first place) , but unless this flag is set, mp3 is ignored from Res as it was before.

hxd/res/Config.hx
```diff
@@ -67,7 +67,9 @@ class Config {
 			Unknown;
 		switch( pf ) {
 		case HL:
+			#if !heaps_enable_hl_mp3
 			ignoredExtensions.set("mp3", true);
+			#end
```
ncannasse: Should mp3 exclude support exclude ogg ? The way we currently deal with it is ignore either one extension or the other, so we can build for either flash (+mp3) or hl (+ogg) and have both mp3 and ogg files in res directory. But that's only for quite old projects and we only use hl/ogg recently.

Yanrishatum: Not particularly. I would expect this flag being used for cases when target platform (flash/html5) supports only mp3, and HL used only for development due to debugger. In such case being able to load mp3 files directly more desirable than mangling 2 formats. 
Predicting "but we can use wav and then convert them" - yes and no. For our particular case, where we use Heaps to develop game on a game jam, being able to just drop mp3 files sent by composer is more convenient than downloading wav files and having them to be converted. 

On a side note, your remarks regarding FLAC made me research flac format a bit and I have WIP flac reader for `format` library. No estimates when it'll be done thought. (Currently I'm able to read the file, but no audio decoding is present and not all metadata blocks are supported)

PR #516, opened 2019-01-01
Add standard cursor override ability.

Yanrishatum:
Introduces `hxd.Cursor.CustomCursor.setCursorOverride` function that allows override of standard cursors. As in: `Default`, `Button`, `Move` and `TextInput`. Custom and Hide cursors cannot be overridden. 
While it's possible to simulate this override manually, it's very hacky and bothersome to do. (setting cursor to each Interactive, using `@:privateAccess` to override TextInput cursor, dealing with Default cursor, etc)
Cursor sample updated to showcase feature as well.

### DIFF
--- a/hxd/Cursor.hx
+++ b/hxd/Cursor.hx
@@ -12,6 +12,25 @@ enum Cursor {
 @:allow(hxd.System)
 class CustomCursor {
 
+	static var cursorOverrides:Map<Cursor, Cursor> = new Map();
+
+	/**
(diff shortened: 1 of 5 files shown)


ncannasse:
This is already supported by the API: you can manually set the variable System.setCursor (which is not a method) and have it call System.setNativeCursor, performing any substitution you might like. 

I agree this might need some sample/documentation update.

Yanrishatum:
Noted. I'll add sample of this api utilization in separate PR.

PR #521, opened 2019-01-10
Change `h2d.Tile` to use Float.

Yanrishatum:
Marked WIP because those are breaking changes, and need a bunch of testing before even considering it being stable.
I need to check lots of other places in codebase to ensure that `Int` -> `Float` conversion gone smoothly.
Then going trough each sample and checking that everything alright.

Overview:
- Changed `Tile` to use Floats on `dx`, `dy`, `x`, `y`, `width` and `height`.
- Changed `Font`, `Text`, `TextInput` and `HtmlText` to operate on Floats to accommodate font resize.
- `h2d.col.Matrix.prependTranslate` - fixed type ambiguity that caused it to resolve to `Int` instead of `Float`.
- `h2d.TileGroup`: Allow using `Float` coordinates.

Detailed changes:
- `h2d.Font`:
  - `baseLine`, `lineHeight`, `offsetX` and `offsetY` changed to `Float`.
  - `setOffset()` now takes `Float` instead of `Int`.
  - `resizeTo()` also resizes Kernings.
- `h2d.Font.Kerning`: `offset` changed to `Float`.
- `h2d.Font.FontChar`:
  - `width`: Changed to `Float`.
  - `clone()`: Fixed incorrect kerning values on resized font by cloning them as well.
- `h2d.Text`:
  - `letterSpacing`, `lineSpacing`, `calcXMin`, `calcYMin`, `calcWidth`, `calcHeight` and `calcSizeHeight` changed to `Float`.
  - `splitText`: `leftMargin` and `afterData` changed to `Float`.
  - `initGlyphs`: `lines` changed to `Array<Float>`.
  - `textHeight` and `textWidth` return `Math.ceil` of `calcHeight` / `calcWidth`. Should be changed to `Float` instead?
- `h2d.HtmlText`:
  - `xPos`, `yPos`, `xMax` and `xMin` changed to `Float`.
  - `sizes` changed to `Array<Float>`.
- `h2d.TextInput`: `cursorX`, `scrollX`, `selectionPos` and `selectionSize` changed to `Float`.
- `hxd.fmt.bfnt.Writer`: Adapt V1 to convert `Font` values to `Int`.
- `hxd.fmt.bfnt.FontParser`: `padding` changed to `Float`.
- `h2d.Console`: `height` calculated as `Math.ceil(font.lineHeight) + 2`.
- `h2d.RenderContext.drawTile`: `calc` function takes `Float` instead of `Int`.
- `h2d.ScaleGrid`: `width` / `height` calculated as `floor(tile.width/height)`. Should be changed to `Float` instead?
- `h2d.Scene.captureBitmap`: `Std.int` on tile params during `setRenderZone`.
- `h2d.CdbLevel`: use `Std.int` on tile height.
- `h2d.filter.*`: Texture allocations use `Std.int(t.width/height)`.

All samples currently compile, but haven't tested yet.

- [x] Initial implementation
- [x] Check all Tile usages and ensure proper adaptation to type switch.
- [x] Resolve questionable places.
  - [x] `h2d.Text.textWidth` and `textHeight` - should they be `Float` or remain `Int`? If later, should they `Std.int` or `Math.ceil`?
  - [x] `h2d.ScaleGrid.width` and `height` - should be `Float` or remain `Int`?
  - [x] `h2d.Interactive.draw` and `getBoundsRec` - should background tile be integer size or subpixel-enabled?
  - [x] `h2d.Tile.fromColor` and other allocations of Tile - should allow subpixel-enabled allocation? ( ceil(size) on texture allocation)
- [x] Test all samples.
  - [x] Flash
  - [x] Hashlink
  - [x] Html5
- [x] Fix discovered issues
  - [x] Fix: Input - Selection only 1 pixel tall
  - [x] Fix: Text rendering broken when `Align != Left`. (Test case: HtmlText/Text)
- Unrelated issues (occurs on Master)
  - Hashlink/JS: filters, Glow - Knockout produces invalid result. (See: https://i.imgur.com/83NI7F4.png HL/JS on the left, flash on the right)
  - JS: filters, Mask - breaks rendering.
  - Flash: s3d shadows are broken.
  - Flash: Pbr - does not disaply the text about unsupported feature.
  - Flash: CubeTexture does not work.
  - Flash: GraphicsDraw - drawPie produces different result.
  - Flash: Helpers are not working.
  - Flash: Lights are not working.
  - Flash: ShaderAdvanced is not working.
  - Flash: Particles2D - in some configurations does not render anything.
  - JS: Fullscreen does not work.
  - JS: PBR have pixel bleeding on skybox. ( see: https://i.imgur.com/yncTnDC.png )
  - JS: Skin - shadows are not rendering properly.


h2d/Scene.hx
```diff
@@ -511,7 +511,7 @@ class Scene extends Layers implements h3d.IDrawable implements hxd.SceneEvents.I
 			target = new Tile(tex,0, 0, width, height);
 		}
 		engine.begin();
-		engine.setRenderZone(target.x, target.y, target.width, target.height);
+		engine.setRenderZone(Std.int(target.x), Std.int(target.y), Std.int(target.width), Std.int(target.height));
```
ncannasse: I guess you should use  x0 = Math.floor(target.x)  , iwidth = Math.ceil(target.x + target.width) - x0

ncannasse:
Regarding your questions, I think they should be Float if we consider all graphics as being subpixel-enabled

h2d/Scene.hx
Yanrishatum: I think about adding `public var ix/iy/iwidth/iheight(get, never):Int` that would return `Std.int/Math.ceil` as a helper properties to Tile. Would save a lot of Float->Int conversion clutter.

ncannasse: I don't think that's best. depending on the condition you might want to floor/ceil

Yanrishatum:
Tested samples, discovered a bunch of issues. Have to check samples with issues on Master to confirm they are related to this PR now, and fix related, probably some of unrelated if able.
Additionally found 2 more places where I'm unsure on how to adapt to type change.

And generally I would suggest bumping library major version when this is merged, because it's a breaking change. Pretty easy to fix, but still.

Yanrishatum:
Checked discovered issues on Master and pretty much all of them are not related to this PR.
To wrap up, need only resolve remaining 2 things:
- `h2d.Interactive.draw` and `getBoundsRec` - should background tile be integer size or subpixel-enabled?
- `h2d.Tile.fromColor` and other allocations of Tile - should allow subpixel-enabled allocation? ( ceil(size) on texture allocation)

ncannasse:
I think if we are alllowing Float tiles sizes they should be enabled everywhere

Yanrishatum:
Done and done. (Also again uikit strikes with build failure. >:( )

h2d/Text.hx
```diff
@@ -256,7 +256,7 @@ class Text extends Drawable {
 
 			if( cc == '\n'.code ) {
 				if( x > xMax ) xMax = x;
-				if( calcLines ) lines.push(x);
+				if( calcLines ) lines.push(Std.int(x));
```
ncannasse: Should be Math.ceil

h2d/Text.hx
```diff
@@ -275,7 +275,7 @@ class Text extends Drawable {
 				prevChar = cc;
 			}
 		}
-		if( calcLines ) lines.push(x);
+		if( calcLines ) lines.push(Std.int(x));
```
ncannasse: Should be Math.ceil

h2d/Tile.hx
```diff
@@ -182,7 +194,7 @@ class Tile {
 	}
 
 
-	public static function fromColor( color : Int, ?width = 1, ?height = 1, ?alpha = 1., ?allocPos : h3d.impl.AllocPos ) : Tile {
+	public static function fromColor( color : Int, ?width = 1., ?height = 1., ?alpha = 1., ?allocPos : h3d.impl.AllocPos ) : Tile {
```
ncannasse: For backward compatibility, we need to keep width : Int and height : Int so that  fromColor(0xFF0000,0.5) - which was color + alpha does not get turned into color + width in a silent manner

h2d/filter/Ambient.hx
```diff
@@ -17,7 +17,7 @@ class Ambient extends AbstractMask {
 	inline function set_invert(v) return pass.shader.maskInvert = v;
 
 	override function draw( ctx : RenderContext, t : h2d.Tile ) {
-		var out = ctx.textures.allocTarget("ambientTmp", t.width, t.height, false);
+		var out = ctx.textures.allocTarget("ambientTmp", hxd.Math.ceil(t.width), hxd.Math.ceil(t.height), false);
```
ncannasse: we should replace all these by a single ctx.textures.allocTileTarget(name,t)

ncannasse:
I think we're good, I'll let @trethaller make a final check and then we can merge. Thanks !

trethaller:
Northgard runs fine, merging now

PR #525, opened 2019-01-16
Flip Cylinder UV coords

lbergman:
Just started using 3D in heaps, so let me know if I'm getting this wrong, but it seems to me like UV coords for cylinder primitive are inverted, resulting in a flipped texture


ncannasse:
@trethaller can you confirm?

trethaller:
Sure, it doesn't seem like we are using addTCoords() anywhere currently

ncannasse:
There was indeed a fix needed, but in the other axis ;)

PR #528, opened 2019-01-19
Fix/improve h2d.Layers

Yanrishatum:
* Added `Object.onOrderChanged` called when changing object ordering in `h2d.Layers`.
* Fixed `h2d.Interactive` not updating it's interaction order when using `over`, `under` or `ysort`.
* Added more features to Layers
  * `getChildLayer` - returns layer on which specified child resides.
  * `getChildLayerIndex` - returns index of a child relative to it's layer.
  * `addAt` - adds child to specific layer and index of that layer. Useful for people who use `Object.addChildAt` and want to utilize adding at index with Layers.
  * `moveChild` - moves Object within layer. Allows for more refined reordering within layer boundaries compared to `over`/`under`.
* Added documentation to functions.

Fixes issue described here: https://community.heaps.io/t/interactives-overlap-layers/160
Edit: Build fails due to PR using failing master build. Has nothing to do with actual PR.

### DIFF
--- a/h2d/Interactive.hx
+++ b/h2d/Interactive.hx
@@ -72,21 +72,15 @@ class Interactive extends Drawable implements hxd.SceneEvents.Interactive {
 		if( backgroundColor != null || forSize ) addBounds(relativeTo, out, 0, 0, Std.int(width), Std.int(height));
 	}
 
-	override function onParentChanged() {
-		super.onParentChanged();
+	override private function onHierarchyMoved(parentChanged:Bool)
+	{
+		super.onHierarchyMoved(parentChanged);
 		if( scene != null ) {
 			scene.removeEventTarget(this);
 			scene.addEventTarget(this);
 		}
-		updateMask();
-	}
-
-	override function onOrderChanged() {
-		super.onOrderChanged();
-		if (scene != null) {
-			scene.removeEventTarget(this);
-			scene.addEventTarget(this);
-		}
+		if ( parentChanged )
+			updateMask();
 	}
 
 	function updateMask() {
--- a/h2d/Interactive.hx
+++ b/h2d/Interactive.hx
@@ -81,6 +81,14 @@ class Interactive extends Drawable implements hxd.SceneEvents.Interactive {
 		updateMask();
 	}
 
+	override function onOrderChanged() {
+		super.onOrderChanged();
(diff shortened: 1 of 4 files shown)


ncannasse:
We already of something in Object to notify of such changes, it's called onParentChanged. Maybe it should be renamed to onParentOrOrderChanged to account for both, I don't think there's a necessity to have two separate yet very similar methods here.

Yanrishatum:
Done. I've kept `onParentChanged` as empty function that called by `onHierarchyMoved` for backwards compatibility sake.
Also fixed oversight that Layers called this function if they weren't allocated.

ncannasse:
I confirm we can remove onParentChanged entirely, it was only used by Interactive to my knowledge and I'm not trying to keep 100% backward compatibility for internals (not public apis). at least not yet.

ncannasse:
Thanks. One last thing I wonder is relative to the functions taking "indexes" (addAt, moveChild, getChildLayerIndex): in Layers, I replaced entirely the notion of child "index" by the notion of "layer", and I find it confusing to reintroduce indexes in the API again. Given both are Ints, one might mistake indexes with layers. Unless you have a rationale for adding these methods?

Yanrishatum:
Those names describe their function - they represent child index on particular layer. Probably worth renaming those as `layerPosition` or `layerIndex` to clarify that those represent child index relative to the layer.
Rationale is pretty simple: They expand `over` and `under` API. They allow for more fine-tuning of object ordering. E.g. "add/move this Object behind/above that Object".
I personally would add even move features, but justifying them would be a hard sell. Ones I had in mind - callback-based analogue to `ysort` (for games that want to use pseudo-Z coord on 2d space), `insertBefore` and `insertAfter` - names are pretty descriptive of function, `changeLayer` to move object between layers (currently it's possible with addChildAt, but it invokes onRemove+onAdd) and `moveLayer` to move layer from index one to another.

ncannasse:
I think that when it comes to doing "advanced" objects management/ordering, you should roll up your own classes either starting from Object or Layers depending on what you need, it's better than using an API that only does a part of what is neeeded.

Since I prefer to keep the API simple for now, please remove these tree index-based methods so I can merge the PR.

ncannasse:
Thanks!

PR #537, opened 2019-01-29
Implement multiple Overs for Interactives

### LINKED ISSUE #209: Over interactive stacking
It should be possible to have several currently "overed" elements when propagateEvents=true

Yanrishatum:
* `currentOver` was replaced by `overList`
* `EOut` events no longer reset `mouseDownButton`, as it prevents firing of `onReleaseOutside`.
* `Interactive` interface expanded to expose `cursor` and `propagateEvents`. Used to properly set cursor when over list changes and do propagation checks.

resolve #209

### DIFF
--- a/hxd/SceneEvents.hx
+++ b/hxd/SceneEvents.hx
@@ -236,8 +236,7 @@ class SceneEvents {
 					overList.remove(overList[idx]);
 					continue;
 				}
-				idx--;
-			} while ( idx > overIndex );
+			} while ( --idx >= overIndex );
 			updateCursor = true;
 		}
 		if ( updateCursor )
--- a/hxd/SceneEvents.hx
+++ b/hxd/SceneEvents.hx
@@ -190,13 +190,6 @@ class SceneEvents {
 							fillOver = i.propagateEvents;
 							overIndex++;
 						}
-						var idx = 0;
-						while ( idx < overList.length ) {
-							var o = overList[idx];
-							if ( o == i ) {
-							}
-							idx++;
-						}
 					}
 				} else {
 					if( checkPush ) {
@@ -304,17 +297,19 @@ class SceneEvents {
 				case EOut:
 					// leave window
 					isOut = true;
-					var i = overList.length - 1;
-					while ( i >= 0 ) {
-						onOut.cancel = false;
-						overList[i].handleEvent(onOut);
-						if ( !onOut.cancel ) {
-							overList.remove(overList[i]);
-							continue;
+					if ( overList.length > 0 ) {
+						var i = overList.length - 1;
+						while ( i >= 0 ) {
+							onOut.cancel = false;
+							overList[i].handleEvent(onOut);
+							if ( !onOut.cancel ) {
+								overList.remove(overList[i]);
+								continue;
+							}
+							i--;
 						}
-						i--;
+						selectCursor();
 					}
-					selectCursor();
 					continue;
 				default:
 				}
--- a/hxd/SceneEvents.hx
+++ b/hxd/SceneEvents.hx
@@ -63,12 +63,8 @@ class SceneEvents {
 	function onRemove(i) {
 		if( i == currentFocus )
 			currentFocus = null;
-		if ( overList.remove(i) ) {
-			if ( overList.length == 0 )
-				hxd.System.setCursor(Default);
-			else 
-				hxd.System.setCursor(overList[0].cursor);
-		}
+		if ( overList.remove(i) )
+			selectCursor();
 		pushList.remove(i);
 	}
 
@@ -125,7 +121,7 @@ class SceneEvents {
 	function emitEvent( event : hxd.Event ) {
 		var oldX = event.relX, oldY = event.relY;
 		var handled = false;
-		var checkOver = false, fillOver = false, checkPush = false, cancelFocus = false;
+		var checkOver = false, fillOver = false, checkPush = false, cancelFocus = false, updateCursor = false;
 		var overIndex : Int = 0;
 		switch( event.kind ) {
 		case EMove, ECheck:
@@ -169,6 +165,7 @@ class SceneEvents {
 								overList.insert(overIndex, i);
 								overIndex++;
 								fillOver = event.propagate;
+								updateCursor = true;
 							}
 							event.kind = oldKind;
 							event.propagate = oldPropagate;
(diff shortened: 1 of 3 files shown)


ncannasse:
> - EOut events no longer reset mouseDownButton, as it prevents firing of onReleaseOutside.

This is meant to prevent triggering onClick.

Yanrishatum:
Since introduction of `onReleaseOutside`, onClick triggers only if mouse is physically over the Interactive. But it does introduce slight change in behavior, so when user presses Interactive, move mouse outside of it and then back inside - currently it will trigger onClick, while previously it wouldn't, which is uncommon for majority of UIs.

Current event chain looks like that when released:
```haxe
onRelease(); // Triggers always and noted in docs.
if (over)
  onClick();
else
  onReleaseOutside();
```
This works, because when releasing mouse outside of Interactive - event no longer disguises itself as `ERelease`, and reports proper `EReleaseOutside`. Which in turn resets `mouseDownButton`.

Yanrishatum:
Bumping. I'm still kind of waiting on this PR. :)

hxd/SceneEvents.hx
```diff
@@ -149,28 +157,49 @@ class SceneEvents {
...
-					if( currentOver != i ) {
-						onOut.cancel = false;
-						if( currentOver != null )
-							currentOver.handleEvent(onOut);
-						if( !onOut.cancel ) {
-							var old = event.propagate;
+					if ( fillOver ) {
+						var idx = 0;
+						while ( idx < overList.length ) {
+							var o = overList[idx];
```
ncannasse: This would be better expressed with `var idx = overList.indexOf(i);`

hxd/SceneEvents.hx
```diff
@@ -200,11 +229,21 @@ class SceneEvents {
...
+					overList[overIndex].handleEvent(onOut);
+					if ( !onOut.cancel ) {
+						overList.remove(overList[overIndex]);
+						continue;
+					}
+					overIndex++;
+				} while ( overIndex < overList.length );
+				if ( overList.length > 0 ) {
+					hxd.System.setCursor(overList[0].cursor);
+				}
```
ncannasse: Two things here:
- I think it's more appropriate to dispatch Out events in the reverse order the Over were dispatched. So since we dispatch Child Over, Parent Over, ParentParent Over, we should dispatch ParentParent Out, ParentOut, Child Out, so iterate in reverse order here
- wha if overList.length == 0 ? We should reset the cursor to Default then

hxd/SceneEvents.hx
```diff
@@ -268,11 +307,15 @@ class SceneEvents {
...
 						onOut.cancel = false;
-						currentOver.handleEvent(onOut);
-						if( !onOut.cancel )
-							currentOver = null;
+						overList[i].handleEvent(onOut);
+						if ( !onOut.cancel ) {
+							overList.remove(overList[i]);
+							continue;
+						}
+						i++;
```
ncannasse: Same remark as above regarding EOut dispatch order

ncannasse:
Added reviews, almost ready for merge :)

hxd/SceneEvents.hx
ncannasse: Minors styling too : please merge the if conditions, such as `if( checkOver && overIndex < overList.length )`

Yanrishatum: When `overList.length == 0` cursor is being reset to Default by `Interactive.handleEvent` .
See here:
https://github.com/HeapsIO/heaps/blob/d1bff99b8f43f2a362fc60af846fcd3637b04d1e/h2d/Interactive.hx#L179-L183
and here:
https://github.com/HeapsIO/heaps/blob/d1bff99b8f43f2a362fc60af846fcd3637b04d1e/h3d/scene/Interactive.hx#L99-L103

ncannasse: I think while we're at it we should let the SceneEvents class handle cursor changes entirely. Actually there's a bugfix to be done as now setting cursor will update it if isOver() while it should only do if we're the "first" (topmost) over with a defined cursor. Maybe calling SceneEvents.cursorUpdated(this) is better

ncannasse: We might also want to minimize cursor changes by only calling setCursor once after all Over's done.

Yanrishatum: In case of `cursor == null`, should it fall trough to the next Interactive in the over list or use `Default`?

ncannasse: I think it should fall through

Yanrishatum: It's still calls `SceneEvents.selectCursor` when cursor is changed on non-top overed cursor, but this is due to fall trough logic, because there may be a case where first Interactive will have `cursor == null` and updated cursor should actually take place.

ncannasse: Makes sense, will review now

hxd/SceneEvents.hx
```diff
@@ -264,16 +299,22 @@ class SceneEvents {
...
-						if( !onOut.cancel )
-							currentOver = null;
+						overList[i].handleEvent(onOut);
+						if ( !onOut.cancel ) {
+							overList.remove(overList[i]);
+							continue;
+						}
+						i--;
 					}
+					selectCursor();
```
ncannasse: only if we had an Over that had a cursor set ?

hxd/SceneEvents.hx
```diff
@@ -149,28 +153,51 @@ class SceneEvents {
...
+							}
+							fillOver = i.propagateEvents;
+							overIndex++;
+						}
+						var idx = 0;
+						while ( idx < overList.length ) {
+							var o = overList[idx];
+							if ( o == i ) {
+							}
+							idx++;
```
ncannasse: Looks like some leftover ;)

hxd/SceneEvents.hx
Yanrishatum: Added check to do it only there was overs present. Still need to call it even if no overs left after cycle, as it would reset cursor to Default in that case.

ncannasse:
Everything seems good on my side, waiting for @trethaller to make a quick test on existing code base before merging.

trethaller:
@ncannasse Northgard looks good with this, except for a crash on EOut, see above

Yanrishatum:
@trethaller Could you elaborate on "see above"? From my limited tests (and I obviously couldn't cover all of them) - I couldn't reproduce the crash.

ncannasse:
There's a comment from Tom in the PR code.

Le mer. 6 mars 2019 19:51, Pavel Alexandrov <notifications@github.com> a
écrit :

Yanrishatum:
@ncannasse Hm. I don't see it? At least I can't find it in Github UI nor I got email notification about it.

trethaller:
@Yanrishatum I opened a review, not sure why you can't see it - anyway I believe the `continue` here shouldn't be there:
https://github.com/Yanrishatum/heaps/blob/feature_propagate/hxd/SceneEvents.hx#L307

Removing it, everything ran fine on my end

Yanrishatum:
@trethaller Hm, that's a weird bug on github side.
Yes, I overlooked this `continue` after reversing EOut call order. Thanks.

ncannasse:
Merged, thanks!

PR #539, opened 2019-02-11
Making sure input material is set

lusius:
During development with GpuParticles we realized addGroup() didn't accept custom material passed and this was needed for having Additive blending.


ncannasse:
@lusius I think it should be more correct to move `if( this.material == null ) this.material = material;` out of the if instead. would that work for you?

lusius:
Agreed. Changed

PR #558, opened 2019-03-09
Implement reference counter on primitives

### LINKED ISSUE #336: Auto-dispose of Primitive
To prevent leaks, an idea would be to add an internal refcount on primitives created in code (eg h3d.prim.Cube) so they get disposed automatically when the last referencing mesh is removed from the scene. 
Primitives coming from loaded models shouldn't be affected by this

Yanrishatum:
Resolves #336 
What I'm not sure about is `dispose` calls. Currently it call dispose on primitive only if it's the only reference to the Mesh.
Adds `refCount:Int` variable, `incref()` and `decref()` methods. Later will dispose of the primitive if refcount reaches 0.
GpuParticles will increase refcounter during serialization to avoid unintentional disposing of it.
HMD models are not disposed due to Library increasing refcounter.

### DIFF
--- a/h3d/parts/GpuParticles.hx
+++ b/h3d/parts/GpuParticles.hx
@@ -966,10 +966,12 @@ class GpuParticles extends h3d.scene.MultiMaterial {
 	override function serialize( ctx : hxbit.Serializer ) {
 		var old = primitive;
 		var oldMat = materials;
+		if ( old != null ) old.incref();
 		primitive = null;
 		materials = [];
 		super.serialize(ctx);
 		primitive = old;
+		if ( old != null ) old.decref();
 		materials = oldMat;
 	}
 	override function customSerialize(ctx:hxbit.Serializer) {
--- a/h3d/prim/Primitive.hx
+++ b/h3d/prim/Primitive.hx
@@ -17,6 +17,12 @@ class Primitive implements hxd.impl.Serializable {
 	public var indexes : Indexes;
 
 	/**
+		Current amount of references to this Primitive.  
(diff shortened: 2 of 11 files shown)


ncannasse:
Regarding dispose() on Object: this calls for an API change. Maybe we should actually remove dispose on Object (2D/3D) then only have it on Scene (2D/3D). Calling dispose() on a scene would be similar to calling onRemove to all objects. 

We might mark the scene as "disposed" to prevent further added objects to call onAdd, and also to eventually call again onAdd if we call render() on it again.

From what I see, this only cause some real difference on h3d.scene.World, but nothing actually bad.

Ping @trethaller for input

ncannasse:
After discussing with @trethaller , I think the good way to go is to entirely remove h3d.scene.Object.dispose (and if the overrides were doing additional things, move them to onRemove),  and keep h3d.scene.Scene.dispose.

Also h3d Scene should have the same behavior regarding onAdd/onRemove as h2d Scene

Yanrishatum:
Removed, except `h3d.scene.World.dispose()`. I'm not sure if it should be relocated to onRemove or kept, as it's disposing not just frees GPU memory but also removes all chunk objects and disposes of textures.

And touching textures. I have a feeling that Textures would benefit from refcounting as well. At least for ones used in h3d materials.

ncannasse:
We might want Terrain Tile to be disposed on remove instead of when manually calling dispose(), I'll let @ShiroSmith and @trethaller look into this particular aspect.

Note for @Fl0xer as well : calling dispose() on s3d no longer auto dispose the World, but I think we're good with NG.

PR #585, opened 2019-04-02
Add SceneEvents.defaultCursor

### LINKED ISSUE #582: h2d.Interactive native cursor bug
h2d.Interactive does not respect hiding of native cursor. When the cursor is on top of it, finger cursor appears and then after the mouse cursor leaves the interactive it changes to pointer and becomes permanently visible. Probably JS specific.

The easiest way to reproduce is to put ```hxd.System.setNativeCursor(Hide);``` at the top of the Interactive2D.hx init() Heaps example. 

The cursor is invisible at first (actually half of the time it's visible and you need to click the canvas to hide it, probably another bug). When you move it over the Haxe logo, the finger appears and when you move the cursor away from the logo, the finger changes to the pointer and stays visible.

Yanrishatum:
Also docs for System.setNativeCursor to clarify it's usage.
Partially resolves #582 
Should there be wiki doc to clarify how to utilize cursor system? 
In general, I thinking about making separate "Community notes" section that will be much more chaotic (code snippets, samples, gists, etc, as name suggests - notes) and faster to produce and not a detailed thought-out doc pages, but they would cover some things community seem to have difficulties with. Lack of documentation is biggest issue people have, and not a lot of people willing to spend a lot of time (including me) to write proper documentation.

### DIFF
--- a/hxd/System.hx
+++ b/hxd/System.hx
@@ -52,6 +52,11 @@ class System {
 	public static function start( callb : Void -> Void ) : Void {
 	}
 
+	/**
+		Sets currently shown cursor.
+		This method designated to be used by custom `hxd.System.setCursor`.
+		Calling it outside of atuomated Interactive cursor update system leads to undefined behavior, and not adviced.
+	**/
 	public static function setNativeCursor( c : Cursor ) : Void {
 	}
 
--- a/hxd/System.hx
+++ b/hxd/System.hx
@@ -54,8 +54,8 @@ class System {
 
 	/**
 		Sets currently shown cursor.
-		This method designated to be used by custom `hxd.System.setCursor`.
-		Calling it outside of atuomated Interactive cursor update system leads to undefined behavior, and not adviced.
+		This method is designated to be used by custom `hxd.System.setCursor`.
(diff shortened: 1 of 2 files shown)


hxd/System.hx
```diff
@@ -52,6 +52,11 @@ class System {
 	public static function start( callb : Void -> Void ) : Void {
 	}
 
+	/**
+		Sets currently shown cursor.
+		This method designated to be used by custom `hxd.System.setCursor`.
```
ncannasse: _is_ designated

hxd/System.hx
```diff
@@ -52,6 +52,11 @@ class System {
 	public static function start( callb : Void -> Void ) : Void {
 	}
 
+	/**
+		Sets currently shown cursor.
+		This method designated to be used by custom `hxd.System.setCursor`.
+		Calling it outside of atuomated Interactive cursor update system leads to undefined behavior, and not adviced.
```
ncannasse: _automated_

ncannasse: advised

ncannasse:
I think a documentation page even if note-based / succinct is better than no documentation at all, so feel free to contribute some rough version. For "community notes", searching https://community.heaps.io/ is already available.

PR #587, opened 2019-04-05
Set public visiblity to init shared shader

D0pu:


ncannasse:
I prefer to keep this private, you can call @:privateAccess sh.initialize() to access it

PR #588, opened 2019-04-05
fix for Bitmapdata.sub() on javascript

infidel-:
From the Mozilla documentation:
```
void ctx.drawImage(image, dx, dy);
void ctx.drawImage(image, dx, dy, dWidth, dHeight);
void ctx.drawImage(image, sx, sy, sWidth, sHeight, dx, dy, dWidth, dHeight);
```
When only x, y arguments are used, they are taken as destination x,y, not source x,y like they should, resulting in black image.

https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/drawImage


ncannasse:
Thanks!

PR #595, opened 2019-04-14
Implement 2D Scene ScaleMode

### LINKED ISSUE #524: [Proposal] Improve `s2d.Scene.setFixedSize`
Given how Heaps garnered interest of 2D devs, it's appropriate to improve some aspects of how scene acts. Currently `setFixedSize` stretches scene contents up to screen size according to internal resolution.
I propose adding `FixedSizeMode` enum, that would include:
- `Stretch` - will behave as now and will be a default.
- `KeepAspect` - will preserve aspect-ratio of the fixed size. Will shift X/Y depending on window size.
Most useful, because most likely people want to keep aspect ratio when screen is resized for majority of games.
- `KeepPixelSize` - same as `KeepAspect`, but will use integer scaling as long as window size > internal resolution.
Will be useful for pixelart games in case developers don't want to stretch pixels (and make them look ugly) even if aspect ratio is preserved.
- (optional) `PreserveArea` - act as `KeepAspect`, but instead of modifying x/y will change width/height to fill entire screen while keeping fixed size "frame".
Avoids letterboxing.

Changes are:
- `public var fixedSizeMode:FixedSizeMode`
- Optional parameter `mode` for `setFixedSize`.
- Actual logic on resize and changes of size mode.

And for clarity, from now on, when I add `[Proposal]` to issue title, I mean that I will implement proposed feature myself if approved.

### LINKED ISSUE #563: Letter/pillarboxing with h2d?
I'm trying to maintain a constant aspect ratio regardless of the window size, and I've been scratching my head trying to figure it out. s2d.setFixedSize(x,y) stretches the image to fit the screen; is there a way to automatically scale the scene so it maintains the aspect ratio and stays centered in the window?

Yanrishatum:
* Closes #524, closes #563.
* Implements 8 different scaling modes for `s2d.Scene`.
* Deprecates `zoom`. Now can be done with `ScaleMode.Zoom(v)`
* Deprecates `setFixedSize`. Now can be done with `ScaleMode.Stretch(w, h)`.
* Removes size checks each frame and instead hooks to window resize events.

Now onto modes:
* `Fill` - it's the default and just resizes Scene to fill engine size.
* `Stretch` - basically good old ugly `setFixedSize`.
* `KeepAspect` - scales Scene with preservation of aspect-ratio. Can anchor on top-left and center.
* `KeepPixelSize` - same as above, but uses integer scaling to ensure square pixels.
* `Exact` - just forces width/height/zoom/centering disregarding adaptation.
* `Zoom` - Same as old `zoom` variable.
* `EnsureArea` - works same as KeepAspect, but instead of having fixed width and height - adaptas it to fill the screen. For example window 800x600 and `EnsureArea(400, 250)` would be upscaled to 2x and cover 800x500, but `height` will be `600`.
* `EnsurePixelArea` - same as above but integer scaling.

There is a particular issue with `drawTile` however - it checks for Scene bounds, not "is it visible on screen". It's logical but may be unexpected for some. We may keep it as is (`scaleMode` docs mention this behavior), force-clip Scene during rendering (`setRenderZone`) or rework how clipper in `drawTile` operates to check window-bounds instead of Scene bounds.

### DIFF
--- a/h2d/RenderContext.hx
+++ b/h2d/RenderContext.hx
@@ -93,7 +93,7 @@ class RenderContext extends h3d.impl.RenderContext {
 		// todo : we might prefer to auto-detect this by running a test and capturing its output
 		baseShader.pixelAlign = #if flash true #else false #end;
 		baseShader.halfPixelInverse.set(0.5 / engine.width, 0.5 / engine.height);
-		baseShader.viewport.set( -scene.width * 0.5, -scene.height * 0.5, 2 / scene.width, -2 * baseFlipY / scene.height);
+		baseShader.viewport.set( -scene.width * 0.5 - scene.offsetX, -scene.height * 0.5 - scene.offsetY, 2 / scene.width * scene.ratioX, -2 * baseFlipY / scene.height * scene.ratioY);
 		baseShader.filterMatrixA.set(1, 0, 0);
 		baseShader.filterMatrixB.set(0, 1, 0);
 		baseShaderList.next = null;
@@ -190,14 +190,33 @@ class RenderContext extends h3d.impl.RenderContext {
 
 		if( restore ) {
 			var tinf = targetsStack[targetsStackIndex - 1];
-			var t = tinf == null ? null : tinf.t;
-			var startX = tinf == null ? 0 : tinf.x;
-			var startY = tinf == null ? 0 : tinf.y;
-			var width = tinf == null ? scene.width : tinf.w;
-			var height = tinf == null ? scene.height : tinf.h;
+			var t : h3d.mat.Texture;
+			var startX : Int, startY : Int, width : Int, height : Int;
+			var ratioX : Float, ratioY : Float, offsetX : Float, offsetY : Float;
+			if ( tinf == null ) {
+				t = null;
+				startX = 0;
+				startY = 0;
+				width = scene.width;
+				height = scene.height;
+				ratioX = scene.ratioX;
+				ratioY = scene.ratioY;
+				offsetX = scene.offsetX;
+				offsetY = scene.offsetY;
+			} else {
+				t = tinf.t;
+				startX = tinf.x;
+				startY = tinf.y;
+				width = tinf.w;
+				height = tinf.h;
+				ratioX = 1;
+				ratioY = 1;
+				offsetX = 0;
+				offsetY = 0;
+			}
 			initShaders(baseShaderList);
 			baseShader.halfPixelInverse.set(0.5 / (t == null ? engine.width : t.width), 0.5 / (t == null ? engine.height : t.height));
-			baseShader.viewport.set( -width * 0.5 - startX, -height * 0.5 - startY, 2 / width, -2 * (t == null ? baseFlipY : targetFlipY) / height);
+			baseShader.viewport.set( -width * 0.5 - startX - offsetX, -height * 0.5 - startY - offsetY, 2 / width * ratioX, -2 * (t == null ? baseFlipY : targetFlipY) / height * ratioY);
 			curX = startX;
 			curY = startY;
 			curWidth = width;
@@ -213,8 +232,8 @@ class RenderContext extends h3d.impl.RenderContext {
 		renderY = y;
 		renderW = w;
 		renderH = h;
-		var scaleX = engine.width / scene.width;
-		var scaleY = engine.height / scene.height;
+		var scaleX = engine.width * scene.ratioX / scene.width;
+		var scaleY = engine.height * scene.ratioY / scene.height;
 		if( inFilter != null ) {
 			var fa = baseShader.filterMatrixA;
 			var fb = baseShader.filterMatrixB;
@@ -228,7 +247,7 @@ class RenderContext extends h3d.impl.RenderContext {
 			w = rx2 - rx1;
 			h = ry2 - ry1;
 		}
-		engine.setRenderZone(Std.int((x - curX) * scaleX + 1e-10), Std.int((y - curY) * scaleY + 1e-10), Std.int(w * scaleX + 1e-10), Std.int(h * scaleY + 1e-10));
+		engine.setRenderZone(Std.int((x - curX + scene.viewportX) * scaleX + 1e-10), Std.int((y - curY + scene.viewportY) * scaleY + 1e-10), Std.int(w * scaleX + 1e-10), Std.int(h * scaleY + 1e-10));
 	}
 
 	public inline function clearRenderZone() {
--- a/h2d/Scene.hx
+++ b/h2d/Scene.hx
@@ -2,19 +2,89 @@ package h2d;
 import hxd.Math;
 
 /**
+	Scaling mode of the 2D Scene.
+**/
+enum ScaleMode
+{
+	/**
+		Fills entire screen. `width` and `height` of Scene will match window size. Default scaling mode.
+	**/
+	Fill;
+	/**
+		Stretches the Scene to fill entire screen. This behavior is same as old `setFixedSize` method.
+	**/
+	Stretch(width : Int, height : Int);
+	/**
+		Preserves aspect-ratio when scaling the Scene.  
+		If `centerViewport` is `true` - Scene will be centered relative to window size.
+	**/
+	KeepAspect(width : Int, height : Int, centerViewport : Bool);
+	/**
+		Same as KeepAspect, but using integer scaling.
+	**/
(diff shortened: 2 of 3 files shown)


trethaller:
@ncannasse I've tested this with the UI scale mode in Northgard and didn't notice any issue. I'm happy to merge it if you agree

ncannasse:
Thanks for the PR and sorry for the review delay.

I'm having some difficulty when I put myself as a user to understand the expected outcome of each scale mode. That's maybe a question of vocabulary or documentation, but let's discuss it before merging.

Fill : I would instead maybe name it as Expand or something, as it enlarge the actual virtual area 

Stretch : ok

KeepAspect : I'm having some trouble with the meaning of width/height in this context. We give an aspect ratio Float (width/height) but then what if the screen does not have the same ratio as the specified one ? In that case when centered we either match the height (thus cutting/expending the left/right borders) or we match the width (thus cutting/expending top/left area).

So I think this should be actually two modes:   KeepWidth( width : Int, centered : Bool ) and KeepHeight( height : Int, centered : Bool )

KeepAspectPixel : I would remove this entirely, as it's very difficult to actually match the good pixel ratio. One way to implement it would be to give the following:
   Pixel( ?minWidth : Int, ?minHeight : Int, ?maxWidth : Int, ?maxHeight : Int );
And the algorithm would try to find the best fixed-size zoom that matches these constraints (if not null) while minimizing constraint breakage if some weird aspect ratio occurs.

Exact : I'm not sure at all what this is doing, it is a stretch with a zoom ?

Zoom : ok. but why would it need to resize "Engine" ? (should actually be named as "screen background buffer")

EnsureArea: from the naming it seems that it can scale down but not up ? and why would it need to resize backgound buffer then?

I guess it's quite difficult to come up with some well described schemes, as we have a scene with no fixed sized that we want to map to a screen with no fixed size either.

There is for example the Unity options for it:
https://docs.unity3d.com/Manual/script-CanvasScaler.html

Actually, once thing that seems important to me, and maybe the source of confusion, is that I think (0,s2d.width)/(0,s2d.height) should always be the _visible_ part of the scene on screen. 

If OTOH we want some kind of fixed size virtual scene which some parts would be actually hidden from screen, I think this would require more discussion about how we represent such area in terms of s2d API.

Yanrishatum:
I admit, my documentation wording and clarity could be better.

Fill: I don't think Expand would convey the function better. Because it adapts virtual size to window size. Expand by meaning would make user assume that it would grow, but will it actually shrink back? Fill on other hand clearly conveys that virtual size will fill entire window size.

KeepAspect: Width/height is here to define the internal virtual size. And when rendering, it will be stretched maintaining aspect ratio of width/height, and ensuring the whole area is visible at all times. It's basically letterboxing mode, so with 4:3 window - 16:9 virtual size will use up all horizontal space and result in letterboxing on top/bottom (or just bottom if centering is disabled). The virtual size remains constant at all times, and what you propose with KeepWidth/KeepHeight is more akin to EnsureArea.

KeepPixelSize: I'm not sure why would that break anything. This mode is basically KeepAspect, but upscale allowed only in integer steps (1x, 2x, 3x, 4x, 5x, etc), this mode is targeted towards pixel-art games which use low internal resolution (e.g. something like 320x240) but need the pixels remain square at all times. With non-integer scaling pixels are not guaranteed to be square. E.g. one pixel is 2x2, but pixel next to it 3x2.

Exact: This is "manual" mode basically. Constant virtual size and visual scale, as well as control on whether center the viewport or not. 

Zoom: Because it's how old `zoom` variable worked. :) 

EnsureArea: I couldn't come up with better name, but as with KeepAspect, it will upscale/downscale the rendered scene, but unlike KeepAspect the virtual size is not constant, but will guaranteed to be minimum of the specified size (e.g. ensured minimum virtual size). 320x200 scene on 800x600 screen would be upscaled 2.5 times. With KeepAspect, it would mean that there would be a letterboxing on top and bottom, because `200 * 2.5 = 500` with 100 pixels out. EnsureArea will include those remaining pixels in the virtual size, resulting in 320x240 virtual size. And it resizes background buffer because for 800x601, there's 1 pixel leftover and uncertain whether to use 320x240 or 320x241, and given that `zoom` solution was to upscale buffer - I did the same here too.

KeepWidth/KeepHeight: As I understood, that would be EnsureArea, but taking into consideration only width or height, which can be achieved with `EnsureArea(width, 0)` and `EnsureArea(height, 0)`

As for Unity:
* Constant Pixel Size = Fill
* Scale With Screen Size = KeepAspect
* Constant Physical Size - no analogue, because scaling mode does not work with DPI.

I suppose I need to work better on documentation explaining each mode? :)

ncannasse:
> I admit, my documentation wording and clarity could be better.
> 
> Fill: I don't think Expand would convey the function better. Because it adapts virtual size to window size. Expand by meaning would make user assume that it would grow, but will it actually shrink back? Fill on other hand clearly conveys that virtual size will fill entire window size.

What about Resize then ?

> KeepAspect: Width/height is here to define the internal virtual size. And when rendering, it will be stretched maintaining aspect ratio of width/height

I my mind, "stretched" != "maintaining aspect ratio"

>, and ensuring the whole area is visible at all times. It's basically letterboxing mode, so with 4:3 window - 16:9 virtual size will use up all horizontal space and result in letterboxing on top/bottom (or just bottom if centering is disabled). The virtual size remains constant at all times, and what you propose with KeepWidth/KeepHeight is more akin to EnsureArea.

So what about naming it LetterBox(width,height) ? I would either drop the centerViewPort entirely or make two enums that tells how to center (the two being optional and default to Center by default) :   LetterBox(width,height,Left|Right|Center,Top|Bottom|Center)

> KeepPixelSize: I'm not sure why would that break anything. This mode is basically KeepAspect, but upscale allowed only in integer steps (1x, 2x, 3x, 4x, 5x, etc), this mode is targeted towards pixel-art games which use low internal resolution (e.g. something like 320x240) but need the pixels remain square at all times. With non-integer scaling pixels are not guaranteed to be square. E.g. one pixel is 2x2, but pixel next to it 3x2.

I see, I still think this should be an option to LetterBox, please change accordingly.

> Exact: This is "manual" mode basically. Constant virtual size and visual scale, as well as control on whether center the viewport or not.

I see. Maybe "Fixed" is better naming then ? And my remarks about centering of LetterBox should be the same here.
 
> Zoom: Because it's how old `zoom` variable worked. :)

Let's slightly break backward compatibility here then : no resize of background buffer ;)
Makes me wonder if we should not change the s2d width/height to Float as we did for most other Ints.
 
> EnsureArea: I couldn't come up with better name, but as with KeepAspect, it will upscale/downscale the rendered scene, but unlike KeepAspect the virtual size is not constant, but will guaranteed to be minimum of the specified size (e.g. ensured minimum virtual size). 320x200 scene on 800x600 screen would be upscaled 2.5 times. With KeepAspect, it would mean that there would be a letterboxing on top and bottom, because `200 * 2.5 = 500` with 100 pixels out. EnsureArea will include those remaining pixels in the virtual size, resulting in 320x240 virtual size. And it resizes background buffer because for 800x601, there's 1 pixel leftover and uncertain whether to use 320x240 or 320x241, and given that `zoom` solution was to upscale buffer - I did the same here too.
> 
> KeepWidth/KeepHeight: As I understood, that would be EnsureArea, but taking into consideration only width or height, which can be achieved with `EnsureArea(width, 0)` and `EnsureArea(height, 0)`

I agree, but I think maybe it should be AutoZoom, and have the width/height optional to reflect the fact that they can be set to 0 to perform auto scaling of one of the dimensions only
 
> As for Unity:
> 
> * Constant Pixel Size = Fill
> * Scale With Screen Size = KeepAspect
> * Constant Physical Size - no analogue, because scaling mode does not work with DPI.
> 
> I suppose I need to work better on documentation explaining each mode? :)

Yes, let's first agree on terminology/API, then please write a bit more documentation for each case.

Yanrishatum:
* Renamed Fill -> Resize
* Renamed `KeepAspect` -> `LetterBox`
* Renamed `EnsureSize` -> `AutoZoom`
* Renamed `Exact` -> `Fixed`
* Unified integer scaling modes with non-integer by adding bool argument to `LetterBox` and `AutoZoom`.
* Added viewport align anchoring for `LetterBox` and `AutoZoom`
* Removed backbuffer resize.
* Added sample showcasing scaling modes.
* `posChanged` is set only if Scene size was changed (won't trigger for constant scene size modes when resizing the window)
* Reworked docs. Hope it's better now.

ncannasse:
Thanks for taking the time to update based on the feedback :)

PR #599, opened 2019-04-16
Fix optim GL unused input shader

D0pu:


ncannasse:
There are several things I'm not sure about there:
- why some WasCleared flag was removed ?
- it is not guaranteed by OpenGL API that the indexes returned by `getAttribLocation` are in the same order as declared in the shader, so I wonder if lastIndex + 1 is ok here.
- in all cases instead of lastIndex , having a currentIndex seems more appropriate
- we sometimes have input with several attributes, and the last Nth ones are not used. I think in that case we should keep previous behavior to prevent binding them.

ncannasse:
What could be done:
- create an Array<Int> initialized with -1 (one per HxSL attribute)
- populate them with getAttribLocation, if >= 0 and keep maxIndex
- when we find "holes" (some -1 at index < maxIndex), allocate an attrib index amount the unused ones (?) this still seems weird to me to actually use a `vertexAttribPointer` on an index that was not returned by `getAttribLocation`

ncannasse:
Why push(false) on curAttribs ? This will make it grow.

Also please remove unrelated change to glslout

PR #601, opened 2019-04-17
WatchFile: Keep reload enabled when parsing crashs

D0pu:


ncannasse:
We don't want to ignore such errors when an error occurs not during reload.

ncannasse:
I don't like it much this way either, as this will prevent us to see the actual call stack when an error occurs in a normal situation

ncannasse:
I don't think times.get(path) == null is good enough to tell if it's a reload, since it might be a file that has been updated but not while reloading in-app. I think it requires passing an optional param to convert(file,?reloading=false) to detect this specific case. 

But then, simply skipping the convert causes another problem, as we will think the file has been converted (according to saved timestamps) while it hasn't... 
So I prefer to reject the solution entirely.

PR #609, opened 2019-05-04
Convert: Use full extension chain

### LINKED ISSUE #607: [Proposal] Use full extension chain for Convert
Currently, `Convert` match only with actual file extensions. While it does the job just fine, at times there is a need for differentiating between same file extensions.
Let me provide an example:
Project contains multiple `json` files, which are named under `<file>.<type>.json` and for each type there's a need for custom Convert. Obviously, making a monster-convert that checks file path and determines what to do is easy, but then there some json files that should not be converted and as far as I'm aware there's no way for Convert to say "Ok, this file should not be converted", hence ability to assign extension chain will be pretty useful. 
On top of that, `Resource` already uses that approach when it comes to `Res` resolving via `Config.extensions`. 
And another real example: We used png files to provide a hit-test mask which had it's own Resource as well as Convert which would process image and generate bitmask of it. Because Convert did not allowed to use `file.hitmask.png` to match, we had to rename actual extension, which was a bit of a pain in the ass, because a lot of graphics editors (namely - Photoshop) are too stupid to read file headers, refusing to open the file just because extension is unknown.

So, proposal:
1. Add `fullExtension` to `hxd.fs.FileEntry` which would return extension chain instead of only last one. (e.g. for file `my.fancy.extended.file` it would return `fancy.extended.file` versus `FileEntry.extension` - `file`)
2. Set convert to use this variable first, and then as fallback to using only actual extension if none matched.

This would preserve old behavior as well as add consistency between Resource and Convert extension handling.
(Honestly, there's more I'd like to say about Convert, but that's out of the scope of this proposal, I'll get to it one day :) )

Yanrishatum:
Closes #607 
* Added `FileEntry.fullExtension` property
* Converts now allow to use extension chains, making it consistent between Resource and Convert.
* Also made `FileEntry.directory` and `FileEntry.extension` to use `indexOf` instead of allocating an array. It hurt my eyes when I've seen this. :)

Now there's another inconsistency btw: `Resource` does not check for real extension, only extension chain while `Convert` does both.

### DIFF
--- a/hxd/fs/FileEntry.hx
+++ b/hxd/fs/FileEntry.hx
@@ -5,7 +5,15 @@ class FileEntry {
 	public var name(default, null) : String;
 	public var path(get, never) : String;
 	public var directory(get, never) : String;
+	/**
+		Real extension of file referenced by FileEntry in lowercase.
+	**/
 	public var extension(get, never) : String;
+	/**
+		Full extension chain of file referenced by FileEntry in lowercase.  
+		For `file.extra.ext` it will return full extension chain - `extra.ext` while `extension` property will return only `ext`.
+	**/
+	public var fullExtension(get, never) : String;
 	public var size(get, never) : Int;
 	public var isDirectory(get, never) : Bool;
 	public var isAvailable(get, never) : Bool;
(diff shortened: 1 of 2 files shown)


ncannasse:
> Now there's another inconsistency btw: Resource does not check for real extension, only extension chain while Convert does both.

Uhm right. For HIDE we allow to detect the type of json based on the "type" value of the top level object. I wonder if we should allow the same or instead use .ext.json

Yanrishatum:
I would say yes. Because file name is easier visible than contents, it also will improve recognition of files when browsing via file browser. Plus HIDE won't have to open the file in order to determine it's type.
And to add up: It also would allow to improve HIDE integration to Heaps, by adding Resources to files like particle configuration jsons (e.g. `file.parts2d.json` / `file.parts3d.json` would get `toParticles()` method that would streamline instancing of them)

ncannasse:
This is somewhat related: I propose we introduce more customizable convert rules, see #613

ncannasse:
Closing (commented on #613)

PR #626, opened 2019-05-31
Add type to App.loadAssets onLoaded argument

Yanrishatum:
Fix vscode `override` completion generating `onLoaded:() -> ?` (which is invalid syntax caused by plugin/completion server not being sure what that function returns).
Also newline because Github build-in editor is "smart". :)

### DIFF
--- a/hxd/App.hx
+++ b/hxd/App.hx
@@ -148,7 +148,7 @@ class App implements h3d.IDrawable {
 		                method when loading is complete
 	**/
 	@:dox(show)
-	function loadAssets( onLoaded ) {
+	function loadAssets( onLoaded : Void->Void ) {
 		onLoaded();
 	}
 


ncannasse:
@Gama11 this should also be addressed by VSCode.

Gama11:
You mean Haxe? Not sure what else to do with a monomorph. :)

Yanrishatum:
Maybe then just generate `onLoaded`, e.g. without explicit type declaration, as it was done in super function?

ncannasse:
not printing the type at all ? That's better than printing an invalid type. And yes haxe should be fixed as well, but it's always better to have a multilayer safety net ;)

Gama11:
>not printing the type at all

Oh, yeah, I guess that would be an option. :)

PR #634, opened 2019-06-21
s2d camera bugfix and zoom added

AxGord:


ncannasse:
@Yanrishatum any insight?

Yanrishatum:
No objections. My 2d camera rework isn't going to be done anytime soon due to workload. So, having bad camera slightly less bad is a good thing.

JarduDray:
So why return scaleX for zoom? Wouldn't getting the value of viewWidth/viewHeight (to get the aspect ratio) multiplied by scaleX be a better option so that everything stays proportional to the screen size, especially when running different resolutions and monitor types?

AxGord:
@JarduDray It is assumed that scaleX == scaleY is always.

ncannasse:
Please reopen the PR, some other PR related to Camera was merged in between.

PR #636, opened 2019-07-01
Make setTransform idempotent

### LINKED ISSUE #635: Idempotency of Object.setTransform
I was playing around with matrix transforms in s3d.scene.object using the [setTransform](https://github.com/HeapsIO/heaps/blob/master/h3d/scene/Object.hx#L772) function.

I noticed that if I called it two times in a row with the same matrix, the scale wasn't consistent.
I noticed that[ it resets the scale of the input matrix](https://github.com/HeapsIO/heaps/blob/master/h3d/scene/Object.hx#L780), for the purpose of extracting the rotation into a quaternion.

To avoid side effects on the input matrix a clone should be made. Or if this is intentional it should be documented

Yann-P:
fixes #635

### DIFF
--- a/h3d/scene/Object.hx
+++ b/h3d/scene/Object.hx
@@ -777,8 +777,9 @@ class Object implements hxd.impl.Serializable {
 		this.scaleX = s.x;
 		this.scaleY = s.y;
 		this.scaleZ = s.z;
-		mat.prependScale(1.0 / s.x, 1.0 / s.y, 1.0 / s.z);
-		qRot.initRotateMatrix(mat);
+		var mTmp = mat.clone();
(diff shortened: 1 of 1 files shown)


ncannasse:
It is correctly inlined / does not perform any heap alloc ?

Yann-P:
I'm not sure about your questions. Do you mean we could make setTransform an inline function if we somehow get rid of the assignment [as stated in the documentation](https://haxe.org/manual/class-field-inline.html)?

trethaller:
thanks @Yann-P, as @Yanrishatum spotted I committed a fix without seeing your PR

Issue #637, opened 2019-07-08
Heaps source breaks on latest Haxe dev Branch from git


JarduDray:
Haxe version: Development branch of git source.
GCC: 9.1.0
OS: Manjaro Linux 64-bit

This error only shows up when building it from source. Latest Binary release is fine:

/home/jardu/haxe/lib/heaps/git/hxd/fmt/hmd/Data.hx:24: characters 18-46 : Abstract 'this' value can only be modified inside an inline function. '_new' modifies 'this'
/home/jardu/haxe/lib/heaps/git/hxd/Res.hx:4: characters 1-8 : Build failure

PR #638, opened 2019-07-08
Fix for bug #637

JarduDray:


ncannasse:
Please do not submit PR that performs reformating additionally to bugfix. You're welcome to resubmit with just the bugfix

PR #646, opened 2019-07-15
Fix JS target after HaxeFoundation/Haxe#8422

Yanrishatum:
HaxeFoundation/Haxe#8422 changed that TypedArrays are no longer compatible with each other directly via `interface`, so there is now a need for implicit declaration that it is indeed an ArrayBufferView. Fixes JS compilation/completion server being unhappy.
P.S. Trusting compiler in figuring out variable type is not always the best decision. ;)


ncannasse:
It's still being discussed on Haxe issues but merging this won't do harm.

Issue #647, opened 2019-07-20
My experiences with h2d.Camera


gordonshamway23:
I experimented a little with the h2d.Camera class and the corresponding example in /samples/Camera2D.hx. I noticed some strange behavior and wonder if this is intended this way.
First of all i did my tests with hashlink sdl or directx build of the sample and started to see what happens if i resize the window. I noticed that after a maximize and a subsequent restore of the window or in general after each resizing of the window the Camera position jumps all over the place and did not behave in a consistent way. So i tried to find out why. I added some changes to the sample in order to better visualize whats going on.
Here is my sample code: "Camera2D.hx"

```haxe
import h2d.TextInput;
import h2d.Interactive;
import h2d.Bitmap;
import h2d.Object;
import h2d.Camera;
import hxd.Timer;
import h3d.mat.Texture;
import h2d.Tile;

class Camera2D extends SampleApp {

	var followObject : h2d.Graphics;
	var camera : Camera;

	var bg : Bitmap;
	var inter : Interactive;

	var sliderAnchorX : h2d.Slider;
	var sliderCamAnchorX : h2d.Slider;
	var sliderCamAnchorY : h2d.Slider;
	var sliderCamViewX : h2d.Slider;
	var sliderCamViewY : h2d.Slider;
	var sliderCamX : h2d.Slider;
	var sliderCamY : h2d.Slider;
	var sliderCamScaleX : h2d.Slider;
	var sliderCamScaleY : h2d.Slider;	

	override function onResize()  {
		super.onResize();


		sliderCamViewX.minValue = -s2d.width*2; sliderCamViewX.maxValue = s2d.width*2;
		sliderCamViewY.minValue = -s2d.height*2; sliderCamViewY.maxValue = s2d.height*2;
		
		sliderCamScaleX.minValue = -s2d.width; sliderCamScaleX.maxValue = s2d.width; 
		sliderCamScaleY.minValue = -s2d.height; sliderCamScaleY.maxValue = s2d.height; 

		updateCamSliderValues();

		bg.tile.scaleToSize(s2d.width,s2d.height);

		inter.x = s2d.width * .5 - 50;
		inter.y = s2d.height * .5 - 20;
	}

	private function setSliderAndTextInputValue(slider : h2d.Slider, value : Float)
	{
		slider.value = value; 
		var  tf : h2d.TextInput = Std.downcast(slider.parent.getChildAt(2),h2d.TextInput);		
[...]

PR #649, opened 2019-07-21
Bugfix in h2d/Camera, correcting calculation of viewX and viewY

gordonshamway23:
Bugfix in h2d/Camera, correcting calculation of viewX and viewY in set_viewX(), set_viewY() and in sync().
Also extend the sample class Camera2D to correctly visualize
camera parameters and the camera anchor point in screen and scene space.
Please see this issue #647 to get further information about the changes.

### DIFF
--- a/h2d/Camera.hx
+++ b/h2d/Camera.hx
@@ -46,11 +46,11 @@ class Camera extends h2d.Object {
 	inline function get_viewY() { return -this.y + anchorHeight; }
 
 	inline function set_viewX( v : Float ) {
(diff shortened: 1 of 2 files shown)


ncannasse:
@Yanrishatum ?

Yanrishatum:
As with another PR - I have nothing to say in particular, since it would be minor fixes for current camera system, while major overhaul is in the works (and specifically because I don't have much time to work on it, It's good to have current system to operate better)

Issue #652, opened 2019-07-27
Heaps build process breaks if resources dir has symlinks in it


infidel-:
If the resources dir contains a symlink, the build process breaks on this line first:
[https://github.com/HeapsIO/heaps/blob/master/hxd/fs/LocalFileSystem.hx#L266](https://github.com/HeapsIO/heaps/blob/master/hxd/fs/LocalFileSystem.hx#L266)
`if( !check || (f == baseDir + path && sys.FileSystem.exists(f) && checkPath(f)) ) {`
because f != baseDir + path. Removing that check works with resources that do not need conversion (though I'm not sure if it's the correct fix).

In case of fonts, though, the process then breaks with font converter in 
[https://github.com/HeapsIO/heaps/blob/master/hxd/fs/FileConverter.hx#L172](https://github.com/HeapsIO/heaps/blob/master/hxd/fs/FileConverter.hx#L172)
`e.file = e.file.substr(baseDir.length);`
because the baseDir has changed (the full filename is now different) and substring will be wrong. Not sure how to fix that.

ncannasse:
Ah, I see, that's the fullPath that follows the symlink...
Actually, we use `f == baseDir + path` to ensure that the file is not accessed with the wrong case on Windows. I guess we could skip this check on Linux that usually does not have case insensitive file system.

One way would be to add:

```
static var isWindows = Sys.systemName() == "Windows";
...
... ((isWindows && f == baseDir + path) && ... )
```

Can you check if that works and if yes submit a PR ? Thanks

infidel-:
Here you go, #653, needed another check. Starting to look a lil ugly :)

PR #653, opened 2019-07-30
symlinks in resources dir fix

infidel-:
Fix for problem one of #652


infidel-:
Sorry, didn't even notice.

Issue #656, opened 2019-08-03
Input normal characters in h2d.TextInput not working on JS target


gordonshamway23:
Hello everyone,

today i found some problems with h2d.TextInput when compiling to Javascript target. I started several samples of heaps containing text inputs and was not able to input normal characters. Using backspace and left and right keys are working. I tested it in chrome and firefox (most recent version) and also with the latest heaps version as well as with the last release of heaps (1.6.1 i think). The problem seems not to be related with heaps. So i debugged a little and maybe found the reason. 
The keypress event is not fired in hxd.Window.js.hx:

```haxe
	function onKeyPress(e:js.html.KeyboardEvent) {
		...
                //actually never called :( 
                // and therefore no ETextInput event is generated
	}
```

The listener is registered in the constructor in hxd.Window.js.hx:
```haxe
              element.addEventListener("keypress", onKeyPress);
```

According to the docs of CanvasElement "keypress" events are deprecated:
[https://developer.mozilla.org/en-US/docs/Web/API/Document/keypress_event](url)

Looks like the event is no longer working in newest browers, even if it's only deprecated, not removed. The docs suggesting using keydown and beforeinput events to replace keypress functionality. I am working on a fix and will propose a PR soon.

Have a nice day.

ncannasse:
As far as I know, keypress events are still correctly supported on canvas with latest browsers. There is no current alternative to keypress as it will handle all shift/alt/ctrl modifiers to translate the key into a proper keyboard-specific character.

If you have no keypress event being fired on the canvas, it means it must be focused first (click on it). Eventually add it a tabindex="0"

gordonshamway23:
Ok, i tested a little more and was not able to get the keypress events correctly working. Neither the tabindex="0" attribute on the canvas element nor clicking the canvas (which i obviously must have done when focusing the TextInput) does help. But i was able to find the actual problem. Calling preventDefault in the keydown eventhandler does prevent the keypress event to fire.

```haxe
	function onKeyDown(e:js.html.KeyboardEvent) {
		var ev = new Event(EKeyDown, mouseX, mouseY);
		ev.keyCode = e.keyCode;
		event(ev);
		if( !propagateKeyEvents ) {
			e.preventDefault(); //this line is actual preventing onKeyPress() 
                                                      // eventhandler being called
			e.stopPropagation();
		}
	}
```

I started a discussion on stackoverflow to confirm that this is intended behavior, and it seems it is.
Please see: 
[...]

PR #657, opened 2019-08-03
Fix for not being able to input characters in h2d.TextInput on JS Target

gordonshamway23:
Please see #656 for an explanation.


ncannasse:
See #656 for discussion.

PR #664, opened 2019-09-03
added: drawRectanglePie()

eminfedar:
I think it is very useful thing.

For example cooldowning abilities you used in a rpg. Ability icons are square, so we need a square pie :)

Screenshot:
Screenshot from 2019-09-04 00-48-58

### DIFF
--- a/h2d/Graphics.hx
+++ b/h2d/Graphics.hx
@@ -545,14 +545,15 @@ class Graphics extends Drawable {
 		flush();
 		lineTo(cx, cy);
 		if( nsegments == 0 )
-			nsegments = Math.ceil(Math.abs(width * angleLength / 4));
+			nsegments = Math.ceil(Math.abs(Math.max(width, height) * angleLength / 4));
 		if( nsegments < 3 ) nsegments = 3;
 		var angle = angleLength / (nsegments - 1);
+		var square2 = Math.sqrt(2);
 		for( i in 0...nsegments ) {
 			var a = i * angle + angleStart;
 
-			var _width = Math.cos(a) * (width/2+1) * 1.42;
-			var _height = Math.sin(a) * (height/2+1) * 1.42;
+			var _width = Math.cos(a) * (width/2+1) * square2;
+			var _height = Math.sin(a) * (height/2+1) * square2;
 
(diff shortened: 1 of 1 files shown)


h2d/Graphics.hx
```diff
@@ -537,6 +537,31 @@ class Graphics extends Drawable {
...
 	}
+	
+	public function drawRectanglePie( cx : Float, cy : Float, width : Float, height : Float, angleStart:Float, angleLength:Float, nsegments = 0 ) {
+		if(Math.abs(angleLength) >= Math.PI*2) {
+			return drawRect(cx-(width/2), cy-(height/2), width, height);
+		}
+		flush();
+		lineTo(cx, cy);
+		if( nsegments == 0 )
+			nsegments = Math.ceil(Math.abs(width * angleLength / 4));
```
ncannasse: should maybe be Math.max(width,height) ?

h2d/Graphics.hx
```diff
@@ -537,6 +537,31 @@ class Graphics extends Drawable {
...
+		if( nsegments < 3 ) nsegments = 3;
+		var angle = angleLength / (nsegments - 1);
+		for( i in 0...nsegments ) {
+			var a = i * angle + angleStart;
+
+			var _width = Math.cos(a) * (width/2+1) * 1.42;
```
ncannasse: Please avoid 1.42 unprecise constant, use (Math.PI * 0.5) instead

eminfedar: Actually it is √2 (1.415) but I didn't want to calculate it everytime because of lacking performance.

But I can make it a constant like this:
`var squareRoot2 = 1.42;`

PI/2 = 1.57, and that makes a latency while changing the angle length.
`1.42` is the best I have experimented :D

h2d/Graphics.hx
eminfedar: yeah absolutely 👍 forgot that.

h2d/Graphics.hx
ncannasse: You can put it into a var outside of the loop :  var sq2 = Math.sqrt(2)

ncannasse:
Thanks!

PR #677, opened 2019-09-28
Reorder EOut/EOver events

### LINKED ISSUE #566: propagateEvents vs overEvent.propagate = true
Discussion following #209 

I was trying to have an interactive with propagateEvents = false but which was setting event.propagate = true in its onOver.

It didn't work very well. First I thought I needed to propagate onMove (and onCheck) as well.

But actually it's a bit more problematic than that, because `fillOver` is checked against `event.propagate` the first time (when we trigger onOver), but then later propogation is only checked with `propagateEvents`.

I guess a more "correct" approach would be to store the result of propagated over. We only need one bool in scene events (the one at the most bottom)

Ping @Yanrishatum

### LINKED ISSUE #675: Over backward compatible behavior
Before #537, the order of events was:
- over element A
- out element A
- over element B

Now it's:
- over element A
- over element B (read propagate result)
- out element A (if propagate = false)

As a result, tooltip systems are slightly more difficult to implement because you need to retain which tooltip belongs to which element in order to properly

Previous behavior:
- over element A : display A tooltip
- out element A : clear tooltip
- over element B : display B tooltip

And also, since this breaks backward compatibility, having a single "singleOver : Bool" flag to turn on to restore previous behavior would be a nice addition.

Yanrishatum:
Resolve #675; Resolve #566
* Ensure EOver is called after EOut
* Remove ability to cancel EOver/EOut
* Update docs for onOver/onOut

For backward compatibility - rel values of Event are retained in case they used during EOver. I'm not sure if `relZ` should be retained. I have a feeling that docs are a bit lacking still in clarifying of EOver/EOut special case.

### DIFF
--- a/h2d/Interactive.hx
+++ b/h2d/Interactive.hx
@@ -249,11 +249,17 @@ class Interactive extends Drawable implements hxd.SceneEvents.Interactive {
 		return scene != null && scene.events != null && @:privateAccess scene.events.currentFocus == this;
 	}
 
-	/** Sent when mouse enters Interactive hitbox area. **/
+	/**
+		Sent when mouse enters Interactive hitbox area.
+		`event.propagate` and `event.cancel` are ignored during `onOver`.
+		Propagation can be set with `onMove` event, as well as cancelling `onMove` will prevent `onOver`.
+	**/
 	public dynamic function onOver( e : hxd.Event ) {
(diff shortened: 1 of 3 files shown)


ncannasse:
Thanks !

PR #680, opened 2019-10-05
Add documentation to `h2d.Layers`

gcardozo123:
Please let me know if I used the wrong code style or anything like that.

### DIFF
--- a/h2d/Layers.hx
+++ b/h2d/Layers.hx
@@ -1,9 +1,7 @@
 package h2d;
 
 /**
- * h2d.Layers allows to hierarchically organize objects on different layers and supports Y-sorting.
- * Note: 
- * 	 If you need a class that supports different scale modes and alignments see h2d.Scene.
+ * `h2d.Layers` allows to hierarchically organize objects on different layers and supports Y-sorting.
  */
 class Layers extends Object {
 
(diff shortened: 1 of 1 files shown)


h2d/Layers.hx
```diff
@@ -1,5 +1,10 @@
 package h2d;
 
+/**
+ * h2d.Layers allows to hierarchically organize objects on different layers and supports Y-sorting.
+ * Note: 
+ * 	 If you need a class that supports different scale modes and alignments see h2d.Scene.
```
ncannasse: I'm not sure I understand this note in the context of h2d.Layers.

h2d/Layers.hx
```diff
@@ -12,14 +17,25 @@ class Layers extends Object {
...
+	/**
+		Adds a child object at the bottom layer (layer 0).
+	**/
 	override function addChild(s) {
 		addChildAt(s, 0);
 	}
 
+	/**
+	 * Shorthand for `addChildAt`, see `h2d.Layers.addChildAt`.
+	 */
```
ncannasse: I think that it should be the other way : add is the correct API function and addChildAt is a substitute. Because addChildAt already has its own definition in h2d.Object, which slightly differs in semantics in h2d.Layers.

h2d/Layers.hx
```diff
@@ -35,6 +51,10 @@ class Layers extends Object {
 			layersIndexes[i]++;
 	}
 
+	/**
+	 * Removes a child `h2d.Object`.
+	 * @param s `h2d.Object` to be removed.
+	 */
```
ncannasse: Given it's an override this does not require documentation (part of h2d.Object)

h2d/Layers.hx
```diff
@@ -12,14 +17,25 @@ class Layers extends Object {
...
 	public inline function add(s, layer) {
 		return addChildAt(s, layer);
 	}
 
+	/**
+	 * Adds a child `h2d.Object` at `layer:Int`.
+	 * @param s `h2d.Object` child to be added.
+	 * @param layer `Int` index of the layer, 0 is the bottom layer.
+	 */
```
ncannasse: No need for documentation here. Should be mentioned in `add`

h2d/Layers.hx
gcardozo123: Removed the note.
You're right, I was trying to teach the user that there is a `Scene` class in case he/she needs, but by the time the user finds out about `h2d.Layers` he/she should know the existance of `h2d.Scene` (because of `s2d:h2d.Scene`).

h2d/Layers.hx
gcardozo123: Fixed

h2d/Layers.hx
gcardozo123: Removed docs that I added on overriden methods.

h2d/Layers.hx
gcardozo123: Fixed.

gcardozo123:
@ncannasse I notice that methods previously documented on `h2d.Layers` weren't showing up on the API page, I think that might be because of missing asteriscs on the beginning of each line, so I added those and took the liberty of slightly changing other method's documentation. Please let me know if any other changes are necessary.

Thanks for the quick review! :)

ncannasse:
Thanks!

PR #683, opened 2019-10-08
Audio system fixes

Yanrishatum:
As per usual, after LD I ended up hating audio system for it's implementation quirks. Time to fix at least some.
* Fixed bug with resampler causing a crash if resampling to higher sample rate (`stb_ogg_sound` on JS crash when using 44100hz ogg resampled to 48000hz)
* Fixed position of the track being broken, because sampleOffset wasn't reset to 0 when stopping source.
* Fixed playback of MP3 on HL. Assuming that mp3 won't have frames with different layers (I couldn't find any samples of such mp3s) - calculations got simplified. Fixed frame offsets calculation.
* Implemented NativeChannel for `hlopenal`. Basically copy-paste of Lime implementation with usage of driver instance. Fixes Sound sample noise not working. Enables code-generated sound on HL in general.
* JS: Implemented `MasterGain`
* JS: Made volume changes on Channel being immediate instead of applied to next buffer fill.
* JS: `Driver.getPlayedSampleCount` no longer returns NaN if AudioContext is in suspended mode during sound startup.
* Improved Sound sample.

I tinkered with current implementation of audio on JS and concluded: It's easier to write custom Driver implementation for WebAudio than doing the Emulator work with it. Would also introduce less overhead.

### DIFF
--- a/hxd/snd/Manager.hx
+++ b/hxd/snd/Manager.hx
@@ -644,6 +644,7 @@ class Manager {
 		}
 
 		while(s.buffers.length > 0) unqueueBuffer(s);
+		s.handle.sampleOffset = 0;
 	}
 
 	var targetRate     : Int;
--- a/hxd/snd/Manager.hx
+++ b/hxd/snd/Manager.hx
@@ -746,7 +746,7 @@ class Manager {
 		}
 
 		if (!checkTargetFormat(data, grp.mono)) {
-			size = samples * targetChannels * Data.formatBytes(targetFormat);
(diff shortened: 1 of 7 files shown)


Yanrishatum:
Okay, before I'll fall asleep. I did implement custom sound Driver around WebAudio, and it works way better than wonky OpenAL emulation layer. Solves frame data overlaying on loops, early cuts (no more hacks with stopping later) and also provides more control over audio chain, allowing for easier addition of features like spatialization (and I already implemented pitch support to a degree). It builds on top of this PR, however and requires some clean up before being ready for showcase.

Yanrishatum:
Followed by #685

PR #685, opened 2019-10-10
WebAudio sound driver and general sound fixes

Yanrishatum:
This PR adds new sound Driver implementation that works directly with Web Audio, providing better sound quality as it skips the AL emulation intermediary and crutches accompanying it.
WebAudio driver is enabled by default unless `useal` flag is set - then it falls back to AL emulator.
Disables buffer resampling to native frequency, since WebAudio can do it internally.
Using direct WA driver also allows implementation of effects, but currently they aren't here yet. As an experiment I implemented 2 drivers: Pitch and LowPass.
For Pitch: Because the time at which buffer going to end changes, there is a need to reschedule future buffers, but it seems I haven't found correct equation to accurately calculate new buffer end time. This results in audio skips  after pitch shift, as well as small, but noticeable gaps on low pitch values.
For LowPass: It almost works. But very first buffer sounds pretty broken, and I'm suspec that's due to how it's implemented in the chain of events:
buffer gets queued (node chain is done) -> effect gets added (chain is rebuilt, but for some reason buffer does not apply it) -> buffer plays without Biquad filter in it's playback chain -> next buffer uses correct chain and works as expected.

There are 3 possible solutions to effects:
* Find a way to switch processing chain mid-playback.
* Find a way to restart currently playing buffer without causing noticeable gap
* Build entire effect chain from the get-go. I'm not so keen about it, as it would produce pretty big overhead for every single source.

This PR incorporates #683 , see it's changelog accordingly.

### DIFF
--- a/hxd/impl/TypedArray.hx
+++ b/hxd/impl/TypedArray.hx
@@ -5,6 +5,7 @@ package hxd.impl;
 #if haxe4
 typedef Float32Array = js.lib.Float32Array;
 typedef Uint16Array = js.lib.Uint16Array;
+typedef Int16Array = js.lib.Int16Array;
 typedef Uint8Array = js.lib.Uint8Array;
 typedef ArrayBuffer = js.lib.ArrayBuffer;
 typedef Uint32Array = js.lib.Uint32Array;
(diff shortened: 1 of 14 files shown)


ncannasse:
- Conflicts when lime removal
- Also, it's a bit of a big chunk, do you think you can split between "js fixes", "mp3 fixes" and "new webaudio driver" ?

Yanrishatum:
I'm not sure. I can try, but my Ludum Dare related break is over already, and I can't say when I'll be able to do it. :)

ncannasse:
Can we close this ?

PR #694, opened 2019-10-21
Added JS support for hxd.File.browse() ("open file" dialog)

deepnight:
Added JS support for the File.browse() method (it uses a temporary <input/> element, but that's how it "works" on JS).


ncannasse:
Thanks !

PR #695, opened 2019-10-21
Fix eager Text rebuild

### LINKED ISSUE #688: h2d.Text updates glyphs very eagerly
```haxe
	function set_text(t : String) {
		var t = t == null ? "null" : t;
		if( t == this.text ) return t;
		this.text = t;
		rebuild();
		return t;
	}

	function rebuild() {
		calcDone = false;
		if( allocated && text != null && font != null ) initGlyphs(text);
		onContentChanged();
	}
```

This causes glyph rebuilding on every text change. When using Text in a way that causes many `set_text` calls (e.g. for a logging component), we end up with a huge overhead (multiple seconds in my case).

I think `rebuild` should instead set some `rebuildGlyph` flag which is then checked by... something else. I tried this in `sync` which seems to work well:

```haxe
	override function sync(ctx:RenderContext) {
		super.sync(ctx);
		if (rebuildGlyphs) {
			initGlyphs(text);
			rebuildGlyphs = false;
		}
	}
```

Not sure if that's the ideal place to check this though.

Yanrishatum:
Fix #688 
Added `glyphsInvalid` to Text, which is set on `rebuild()` call, and unset only when glyphs are rebuilt. The only place where it's required to have glyphs present is `draw` (all other metrics already use `calcDone` and calculate virtual size), so it's checked only here.
I'm not too happy with that solution, to be honest. First, it causing glyph initialization on `draw` call, while it probably should be done in `sync`. Second, extra variable that basically same as `calcDone`. Maybe change it to `abstract(Int)` to hold both flags in one variable (+ getter/setter for backward compat)?


ncannasse:
We don't want to perform initGlyphs during draw() if this can be avoided, it should be done in sync() before that, as this is much better in JS for example to first upload all the buffers and then perform draw calls

Yanrishatum:
I've kept check in `draw`, because when rendering to texture, it's not guaranteed to call `sync` as far as I remember.

ncannasse:
Another thing I'm thinking while we're at it.
Doing:

```haxe
tf.text = "abc";
tf.text += "d";
```

will rebuild the glyphs at every frame instead of caching the previous result.

Yanrishatum:
As you mentioned in comment for splitText - that calls for major refactoring. That would be quite trivial if done with approach like OpenFL for example (e.g. it internally splits text to chunk instances which have their own format, and appending text would rebuild only last chunk or generate new one, it also would make HtmlText share implementation with Text and just would allow for more control over format, and also programmatic HtmlText construction without XML).
Right now caching previous result would be problematic for aligns other than Left, because appending means that there have to be recalculation of entire line, which is impossible with current implementation. Mainly because `TileGroup` does not allow altering of glyph data after they were set.
First step to allow text append is to allow altering existing quads in TileGroup.

ncannasse:
My idea was just to differentiate a text change from other changes requiring full rebuild, and keep a variable that says which was the last text rendered, so we can early exit if no rebuild is required

Yanrishatum:
I'm not sure I understood you. `rebuild` is called when `text`, `textAlign`, `letterSpacing`, `lineSpacing` or `maxWidth` are changed, `onAdd` (e.g. allocation) happens or constraint update with `updateConstraint`, which happens with Flow or changing `maxWidth`. In all of those cases visual appearance of Text changes, requiring full rebuild. Because `TileGroup` cannot be changed once you sent data to it - there isn't a way to just rebuild small part of it. Appending text with align other than Left guarantees to change rendering of entire line, and even with Left, word-wrap can cause 1 to 2 lines requiring rebuild. So yeah, due to that specific behavior there's no way for a partial rebuild in current implementation. PR solution marks glyphs as invalid, and `initGlyphs` happens only if text dimensions are queried or `sync` / `draw` happens. But glyphs are rebuilt only on last 2, but not on dimension query.

ncannasse:
My suggestion was the following:
- have two flags "needRebuild" , that is trigger by all major changes as currently done and "checkText" that is only triggered by text content change
- on sync, we check  `if( checkText && this.text != currentText ) needRebuild = true;`
- on rebuild, we do `needRebuild = false; checkText = false; currentText = text;`

This way doing :   tf.text = "a"; tf.text += "b"; every frame should not trigger text rebuild every frame.

Yanrishatum:
Now I understand. 
Now `set_text` only sets `textChanged` flag and triggers `onContentChanged` (see below). I've kept `calcDone` and `needsRebuild` separate, however, because otherwise there would be a need to update glyph cache on accessing size metrics as well (or perform metric calculation on each access until glyphs are rebuilt). In current implementation it is mitigated by the fact that glyphs are rebuilt instantly on any field change that affects display, e.g. eagerly.

As for why `onContentChanged` triggers regardless: Short: Because Flow.
Long:
1. Text gets changed (only text, not any other text-display affecting field)
2. Flow.sync()
    * If reflow - reflow happens with old metrics, because text was changed, but flag to remake calculations wasn't.
    * If not - flow breaks, because new text metrics won't be accurate
    * In any case - `needsReflow` flag would never be set by Text before `sync()`, guaranteed causing 1-frame delay of reflow with new metrics.
3. Flow causes `Text.sync()`
    * Text updates it's metrics. Flow is unaware of that.
    * If it and calls `onContentChanged()` it would be too late and Flow would reflow only next frame.

As to why `rebuild` does not do `initGlyphs` - since main issue was eager rebuild on Text - all other fields did it eagerly too. E.g. in case of that:
```
// We assume that s2d is allocated and causing onAdd.
var text = new h2d.Text(font, s2d); // onAdd() -> rebuild();
text.text = "Some text";
text.align = Center; // rebuild();
text.letterSpacing = 0; // rebuild();
text.lineSpacing = 5; // rebuild();
```
Oh my, 4 whole rebuilds with `initGlyphs()` instead of one, and 3 of them are with actual text!
If we do lazy rebuild, better go all the way ;)

Yanrishatum:
* Implemented validation and precaching in HtmlText. Now eagerly parses Xml and fetches all img tiles as well as fonts. I'm not sure if I should reset `textXml`

ncannasse:
I don't think we need a cache at a per-textfield level. That's quite inefficient and if there are too many requests a global cache implemented by user would get much better results and correctly handle invalidation where it matters (for instance switching resolution might switch some fonts).
So I would simply call loadFont etc. at validation time.

Yanrishatum:
While I understand where this coming from, but caches are here to ensure consistency. So for example when using `<font face="random">` that would return random font it would not cause broken metrics, same goes for `` - it's not guaranteed that `loadFont` and `loadImg` would be deterministic, and return same values over 3 steps of building HtmlText (validate -> build metrics -> build visuals). Now that I think about - my own sample shown a specific bug - `` would show same image, as `random` is cached. (Can be mitigated with using linear Array, since nodes are processed linearly anyway).
Alternatively, I can remove the image/font caches and just add note to documentation that says that functions have to be deterministic as to not break the rendering.

ncannasse:
>Alternatively, I can remove the image/font caches and just add note to documentation that says that functions have to be deterministic as to not break the rendering.

That seems good enough.

PR #696, opened 2019-10-23
Added compiler flag for disabling window resize

jefvel:
Compiling to hl using hldx or hlsdl with `-D windowDisableResize`
will create a window that is not resizable.


jefvel:
oh, this is exactly the same as 678. Closing this.

PR #697, opened 2019-10-24
Fix `h2d.Console` scaling

gcardozo123:
This fix is necessary to prevent the console from going out of the screen after being scaled up.

Here's a little [test project](https://github.com/HeapsIO/heaps/files/3765198/console_scaling_test.zip). Notice that I changed the console key in the project to `'` because it's easier for the Brazilian keyboard.

### DIFF
--- a/h2d/Console.hx
+++ b/h2d/Console.hx
@@ -423,9 +423,8 @@ class Console #if !macro extends h2d.Object #end {
 		var scene = ctx.scene;
 		if( scene != null ) {
 			x = 0;
-			y = scene.height - height;
-			width = scene.width;
-			tf.maxWidth = width;
+			y = (scene.height / scene.matD) - height * scaleY;
+			width = Math.ceil(scene.width / scene.matA); 
 			bg.tile.scaleToSize(width, height);
 		}
 		var log = logTxt;


ncannasse:
A better fix would be to use the calculated object matrix to also take parents into account. And it also requires dealing with width as well.

gcardozo123:
@ncannasse I amended my commit, I'm not sure if this is what you meant but it worked. I tested setting different scales for both `s2d` and the `console`.
I didn't had time to check the `width` but during my tests it never became smaller than the width of the window.

I

gcardozo123:
Wow, looks like the console doesn't handle its width even if you don't change the scale of anything.
I changed it's `y` position just to check what happens if the text takes more space than the screen width and here it is:
console-width-bug
I purpose we create an issue for the `width` as it seems to be a separate problem than the one I'm trying to fix.

ncannasse:
I think the two are related and should be fixed together. Console was not designed with scaling in mind.

gcardozo123:
> I think the two are related and should be fixed together. Console was not designed with scaling in mind.

Fair enough. 
Taking a quick look at `h2d.Text` and `h2d.TextInput` it seems that they don't have a "single line" mode. Do you think we should add modes to `h2d.TextInput` as an `enum` or just use a `Bool` [like OpenFL does](https://api.openfl.org/openfl/text/TextField.html#multiline)?

h2d/Console.hx
```diff
@@ -423,9 +423,8 @@ class Console #if !macro extends h2d.Object #end {
 		var scene = ctx.scene;
 		if( scene != null ) {
 			x = 0;
-			y = scene.height - height;
-			width = scene.width;
-			tf.maxWidth = width;
```
gcardozo123: I removed this line because setting `tf.maxWidth` makes `TextInput` become multiline, is that intented or should I open an issue?

qkdreyer (approved):

ncannasse:
I don't think the fix is correct as it should take into account not the scene but the console local transform (as the console might be scaled or either put into a scaled container).

PR #698, opened 2019-10-24
Implement new s2d camera

### LINKED ISSUE #622: [Proposal] Implement s2d cameras
First paragraph is premise as usual, propsal afterwards.
I know, I know, it's already was discussed before, and Cannasse was adamantly against the idea of using camera transform inside shaders, and insisted that having it as root object tree is the way to go. I still disagree and still insist that it's a stupid idea. Yet my solution back then was not a proper solution and I admit it. This is basically a take 2 on that matter, that is more researched and will be based on #595 .

Proposal: Introduce GPU-side 2D camera transformation by changing `viewport` variable to `viewportA` / `viewportB` matrix.
* Instead of #595 calculating offsets and scaling relative to screen, Scene will contain `a`, `d`, `tx` and `ty` transform variables that would convert scene coordinate space to -1...1 screen space. (`b` and `c` are omitted, because they are always `0`)
* Base2d shader will apply `dot(viewportA)` / `dot(viewportB)` matrix instead of `(pos.xy + viewport.xy) * viewport.zw`. It does increase calculation cost slightly, however.
* In `RenderContext`: `curX`, `curY`, `curWidth` and `curHeight` will be replaced by `viewportMatrix` (either as separate variables or `h2d.col.Matrix` instance).
* `RenderContext.drawTile` will always apply viewport matrix and determine if tile is inside the screen by -1...1 results. It is more costly than current approach, however.
That also will solve issue present in #595 that causes Tiles being clipped out, if they are outside of virtual scene viewport, but still inside window visible area.

Now onto implementing a better camera system. I think of two primary paths to take:
1. Add `h2d.Scene.camera` variable, which would dictate additional viewport changes apart from scene viewport.
   * By default camera is either `null` or one that does not perform any extra modifications resulting in backward compatibility of how currently s2d operates.
   * This approach will make it consistent between s2d and s3d, as both scenes will contain a camera which can be manipulated.
   * Downside: Harder to setup multiple cameras (use case: split-screens, picture-in-picture)
   * Downside: Makes it more problematic to implement something like top-most UI and scrolling gameplay area.
2. Rework `h2d.Camera` as a container for one or multiple Camera instances.
   * Camera transforms applied only if actual Camera instance is involved.
   * Closer to how Camera works now, except does not invalidate entire goddamn object tree positions.
   * No downsides from above.

I lean towards second approach. I still hope to implement capability to have multiple cameras (ones that do not require duplicating object tree), but it's most likely something that will be available via API and not as easy to set up premade class. (Or make those premade classes as not part of stock Heaps. They are very situational, and not everyone needs multiple cameras, however just one camera that just works is important.)
As for camera API:
`RenderContext.setCamera(cam)` and `RenderContext.resetCamera()` that would affect viewport matrix. Simple, but there isn't much to add, really.
Modifications also will affect `pushTarget` and `popTarget`, as they set the shader viewport.
Also I will need to test how well it handles filters, as I'm not sure how it would handle it.

Summary of downsides:
* More costly `RenderContext.drawTile`
* Slightly more costly vertex shader for s2d.
* Breaks compatibility with current `h2d.Camera` in any case.

Upsides:
* Coherent and actually usable camera system.
* Setting up multiple cameras is actually possible without duplicating entire object tree.
* Moving camera does not invalidate positions of entire object tree residing in camera.
* Possible to perform complex transformations with rendered content via camera at no cost increase.
* `RenderContext.drawTile` won't have clipping "bug" as described in #595 
* People finally will stop complaining about camera in Discord chat :) 
* I will stress it again: Moving camera won't trigger position invalidation in object tree. And should't in the first place.

I already have a dirty WIP based on PR mentioned 4 times already, which implements the changes before adding camera transform into equation and so far it works well.
RFC, I guess?

Yanrishatum:
Or, in simple words: Camera that does not suck. :D 

## Implementation details
As Cannasse proposed - I went with Layers being camera containers. Each Layers contains at least one camera instance that renders everything by default. I kinda accidentally created full-on multi-camera support with that as well. 

* Changes to shaders
  * Now uses transformation matrix to convert world space to screen space.
* Changes to camera:
  * No longer an h2d.Object - is a breaking change, but I pity anyone who had to use initial h2d.Camera.
  * Allows to clip out out-of-bounds rendering with `setRenderZone` and flags.
  * Works relative to scene boundaries dictated by `scaleMode`.
  * Allows to control "anchor" position of the camera, defaulting on top-left corner. Saves a lot of headache when scene size changes, as you just need to change anchor.
  * Allows for scaling and rotation. (with origin being anchor)
  * Allows to filter out layers. For example, to make effortless static UI - you can just add second camera, and tell first not to render UI layer, and second to only render it. I swear I made multi-camera support accidentally :D 
* Changes to Layers
  * Now contain cameras array, and renders accordingly.
  * Initializes with one camera by default.
  * Also provides shortcut to first camera with `camera` property.
* Changes to Object
  * Clipping math in screen space for filters.
* Changes to Scene
  * Changed `scaleMode` math as to calculate A and D components of world-to-screen matrix and X/Y offsets being in screen-space.
* Changes to RenderContext
  * Thrown curX/Y/W/H out of the window - welcome view matrix.
  * Use new view matrix everywhere.
  * Added `setCanera` / `resetCamera`
  * Removed `restore` flag from `popTarget`, because I couldn't find any place that uses it, and it's pretty dangerous thing in itself.

## Considerations
The only major thing to implement that remains is Interactives, and by god I have no idea where to start. So I think about slight shift in design. I want to remove cameras from Layers and instead implement it inside Scene. That would automatically solve camera-in-camera shenanigans and camera-in-filter too. BUT it will disable ability to set filters to Scene, at least until cameras won't allow for individual shaders. That would make implementation of Interactives far more easier.

Also cameras are not optional now. Using them totally is, but at least one camera would always be present. Making them opt-in to a level that if you don't need it - there isn't one at all would overcomplicate entire rendering logic, not to mention Interactives. It isn't that big of an overhead in general.

## TODO

- [x] Core camera
- [x] `drawTile` clipping
Transforms tile position to screen space (-1...1) and checks against that.
- [x] Ensure filters are working inside cameras
  * Bug: When camera is rotated, masks act out. I'm not sure if I can fix that. May have more issues with masks.
- [x] `pushRenderTarget` / `popRenderTarget`
Reworked stack slightly to restore state completely and not rely on scene data (because cameras)
- [x] `setRenderZone`
- [x] Camera resync on window resize
I just force sync when position of camera container is changed, which happens on window resize.
- [x] Half-pixel issues due to different `scaleMode` calc.
- [x] Allow camera to clip out contents with `setRenderZone` (`viewport` variables)
It deals with scene-size relative coordinates here.
- [x] Move cameras from Layers to Scene
- [x] Deal with Interactives inside camras.
- [x] Code cleanup
- [x] Documentation
- [x] Camera2D sample

UPD: Also closes #622


Yanrishatum:
* Exactly one camera can work with Interactives. Allowing multiple cameras to do it would require large amount of reworks as well as breaking changes (for example `mouseX/Y` would have to be moved to Camera), so I decided that it's better to just enforce one interactive camera. That should cover majority of use-cases with mouse interaction. It can be assigned freely, because user would probably want to set it to UI camera, which would be among the last in the list (due to rendering order).
* There is a known bug that if you assign filter to Scene - camera system would break. That is due to `drawFilters` not using cameras, as I want to avoid copy-pasting rendering code. I already did it with `drawRec` anyway.
* Another bug is Mask filter - with rotated camera it sometimes does not accurately masks the content.
* Added very primitive ability of Camera to follow objects. 

Induced rendering overhead:
Yes, it's obviously here, and it's mainly extra matrix transformations performed during tile rendering, filter rendering and camera enter phase. (Last part applies Scene multiplies camera transform onto Scene viewport)
Also calculation of mouseX/mouseY got a tad more complex, because of rotations in the camera.

Questionable:
* I'm not sure if I should convert coordinate passed to `Scene.getInteractive` to camera space or not.
* I can fix the `drawFilters` issue with slightly changing `drawRec` and `drawFilters` to call `drawSelf` which would perform actual rendering (e.g. `for (child in children) child.drawRec(); draw();`). But it would change rendering of filters a bit, because it would start to respect `front2back` during rendering.
* I'll probably omit it in this PR - but I'll investigate ability to add filters to cameras. That would be pretty useful when applying post-process effects to game (e.g. blur, classic "underwater" effects, etc) that should not affect UI or other cameras.
Opinions? @ncannasse

With that, primary implementation is complete, I still need to check for code style, perform clean ups, update 2D camera sample and resolve questions above.

ncannasse:
The one single camera for Interactives limitation is not good. You might
want clicks on both your world and your UI.

Le dim. 27 oct. 2019 à 07:22, Pavel Alexandrov <notifications@github.com> a
écrit :

ncannasse:
And the coordinates are always in Interactive coordinates, not in camera
ones. Camera is for display, it should not affect each object individual
coordinates or events position.

Le dim. 27 oct. 2019 à 08:09, Nicolas Cannasse <ncannasse@gmail.com> a
écrit :

Yanrishatum:
> The one single camera for Interactives limitation is not good. You might want clicks on both your world and your UI.

Agreed, but as I mentioned - it would require a lot of rework and breaking changes. It is a thing that worth adding to plans on adding, but I honestly not up to it for now. One handler covers a lot of use-cases already, and it's not like it reduces feature that, because there were no multiple cameras to begin with. And while it's not possible, there are workaround. (For example - using just one camera and sync UI object position to camera, so it appears in same place, but moving in reality. Still 100 times better than invalidating entire object tree when moving camera - just UI part)

> And the coordinates are always in Interactive coordinates, not in camera
ones. Camera is for display, it should not affect each object individual
coordinates or events position.

It still receive those coordinates in Interactive local coordinates. What I meant is - should `Scene.getInteractive` use global coordinates when hit-testing or camera coordinates. Since global coords and camera coords are now detached. But it seems it should work with global coords, as it's not mouse interaction per se.

Yanrishatum:
* Redid Camera2D sample to showcase new camera system.
* Added `followRotation` flag to camera - allows to sync up rotation of camera with rotation of followed object. (showcased in the sample)
* Adde `setViewport` helper method as a shortcut to setting camera viewport boundaries.

With that, PR is ready for review. Other features, like filter or shader support on cameras and support for interactives on multiple cameras requires further consideration and outside of the scope of initial camera system implementation.

Yanrishatum:
4 weeks anniversary bump 🎉
@ncannasse ;)

ncannasse:
I don't forget about it 😉

Le dim. 24 nov. 2019 à 19:32, Pavel Alexandrov <notifications@github.com> a
écrit :

ncannasse:
Ok i did a first review pass, here's the comments:

- setCamera/resetCamera should be renamed pushCamera/popCamera as it operates on a stack
- make the CameraStackEntry and TargetStackEntry private

- in pushTarget (and pushCamera?), given there's a lot of vars now written I would prevent the duplication, and change:

```haxe
		if ( targetsStackIndex > targetsStack.length ) {
			targetsStack.push( {
				t: curTarget,
				va: viewA, vb: viewB,
				vc: viewC, vd: viewD,
				vx: viewX, vy: viewY,
				hasRZ: hasRenderZone,
				rzX: renderX, rzY: renderY,
				rzW: renderW, rzH: renderH
			} );
		} else {
			var o = targetsStack[targetsStackIndex - 1];
			o.t = curTarget;
			o.va = viewA;
			o.vb = viewB;
			o.vc = viewC;
			o.vd = viewD;
			o.vx = viewX;
			o.vy = viewY;
			o.hasRZ = hasRenderZone;
			o.rzX = renderX;
			o.rzY = renderY;
			o.rzW = renderW;
			o.rzH = renderH;
		}
```

into 

```haxe
var t =  targetsStack[targetsStackIndex - 1];
if( t == null ) { t = { ... zeroes }; targetStacks.push(t); }
... assigns
```

- I haven't fully checked the clipping check for correctness. I _hope_ it's correct :)

- do we really need the viewport properties of Scene accessible in public ? Given they are in screen space now, unless necessary we should keep them private

- what about filters on Scene ? it seems to entirely ignore camera. Maybe camera should have filter instead, and scene filter should apply to default camera (throw error if both have a filter)

- " Scalid applied relative to anchored position." : there's a typo (Scaling) and also "relative to anchored position" is a bit unclear.  "Scaling, using anchored position as pivot" seems more accurate maybe?

- I'm not so much happy with camera viewport properties being in [0-1] range. Maybe instead these should be private properties, and public API should always multiply/divide by current scene size instead ?

- why expose both "point" apis and "X/Y coordinate" apis for Camera ? I think point is enough. Other should be made private if used by Scene for example.

- just an idea, haven't thought at all the implications, but maybe we should separate Scene from CameraView.  Scene would still extend CameraView that would extend Layers. The idea is to be able to have a CameraView inside the Scene but not a the topmost level. Does it make sense?

Otherwise, once these details solved, it's good to merge.

Yanrishatum:
Sorry for delayed answer (there are reasons ;) ) 

> push/popCamera, private typedefs, typos
Will do.

> given there's a lot of vars now written I would prevent the duplication
Is there any benefit in doing so? Just curious. (Apart from more code points to maintain if it would change in future) Allocation shouldn't happen that often (e.g. once when new cameras added), so it's logical change for maintainability reason.

> do we really need the viewport properties of Scene accessible in public ?
Not much, no. I'm pretty sure it can be safely set private. If anything, people can use `@:privateAccess` to get those out.

> what about filters on Scene ? 
Yes, currently filters on whole scene level are not implemented. Only use-case I can think of are post-processing filters, but I wasn't sure how to deal with those. Main problem is how Object with filters renders onto Scene, as it uses completely detached method when drawing children, which also does not follow some rules, such as front2back. To enable ability to apply camera/scene-wide filters, there is a need to override `drawFilters` and pretty much copy-paste entire routine just to add camera handling. Alternatively I can extract the `draw(); ... child.drawRec()` part in separate function and use common code for both. But that still raises question - should it follow front2back rule or not when drawing filters. As for `filter` on Scene - I would just disable it outright to keep it simple and allow it only on cameras.

> Maybe instead these should be private properties, and public API should always multiply/divide by current scene size instead ?
In that case it would lead to bugs when preparing multiple scenes in advance, as well as unpredictable values when dealing with window resizes. (set one value -> window got resized -> now have to set different value based off current scene size to get same result)
Additionally it introduces dependency on current scene (And I have no clue how to actually get "current scene" without having direct reference to it, `hxd.App` is not exposed anywhere as public variable/getter), while rest of the Camera is designed in a way that it does not care about which Scene tries to render it, and when it does need Scene values - Scene instance is passed to Camera as an argument.
Introducing methods that allows to set viewport in pixels relative to given Scene is a good idea, however.
Alternatively - abandon all independent camera thing and enforce that it can have only one Scene tied to it, but that would require to change API a bit to ensure user won't screw up adding/removing cameras. (One of reasons for independent camera is ease of use - you just `new Camera()` and then push it to `Scene.cameras`)

> why expose both "point" apis and "X/Y coordinate" apis for Camera ?
Mainly for more flexible API reasons. For example when trying to convert x/y, but without need to allocate a Point instance. If anything, I think just hiding them behind `@:noCompletion` would do the trick?

> just an idea, haven't thought at all the implications, but maybe we should separate Scene from CameraView
It would pretty much same as making Layers as host of the Camera. I initially went with Layers, but when it came to handling Interactives, I soon realized that having camera anywhere not on top-most level would cause a lot of problems. Can we be sure if Interactive is inside camera or not? Should be traverse back on it's parent tree just to check each time we hit-test? What if it's camera inside another camera? While camera internally is a stack, I don't want to allow camera-in-camera behavior as it complicates everything once again (but if anyone wants - they absolutely can mess with camera-in-camera, it's just I don't want to hold responsibility if anything goes wrong :) ). On top of that, having it on Scene makes it easier to understand. 3D scene have camera, 2D scene have camera. They are different, but main concept of Scene having a camera remains same.
What is the benefit of going trough all the trouble of setting up "if there's CameraView, then do A otherwise do B"? I understand that having it completely optional to the point of not doing extra calculations associated with it is compelling, but it simply introduces a lot of complexity and makes it way harder to maintain.

ncannasse:
> Sorry for delayed answer (there are reasons ;) )
> 
> > push/popCamera, private typedefs, typos
> 
> Will do.

Tell me when ready

> > given there's a lot of vars now written I would prevent the duplication
> 
> Is there any benefit in doing so? Just curious. (Apart from more code points to maintain if it would change in future) Allocation shouldn't happen that often (e.g. once when new cameras added), so it's logical change for maintainability reason.

It's more concise code style, and ensure that if we change something we don't have to do it in the two code paths.
 
> > do we really need the viewport properties of Scene accessible in public ?
> 
> Not much, no. I'm pretty sure it can be safely set private. If anything, people can use `@:privateAccess` to get those out.

Better, I prefer to keep the API public to only what's meant to be used by end-user.
 
> > what about filters on Scene ?
> 
> Yes, currently filters on whole scene level are not implemented. Only use-case I can think of are post-processing filters, but I wasn't sure how to deal with those. Main problem is how Object with filters renders onto Scene, as it uses completely detached method when drawing children, which also does not follow some rules, such as front2back. To enable ability to apply camera/scene-wide filters, there is a need to override `drawFilters` and pretty much copy-paste entire routine just to add camera handling. Alternatively I can extract the `draw(); ... child.drawRec()` part in separate function and use common code for both. But that still raises question - should it follow front2back rule or not when drawing filters. As for `filter` on Scene - I would just disable it outright to keep it simple and allow it only on cameras.

We need to have the filter on scene working some way or another, especially when there's only a single (default) camera set.

> > Maybe instead these should be private properties, and public API should always multiply/divide by current scene size instead ?
> 
> Alternatively - abandon all independent camera thing and enforce that it can have only one Scene tied to it, but that would require to change API a bit to ensure user won't screw up adding/removing cameras. (One of reasons for independent camera is ease of use - you just `new Camera()` and then push it to `Scene.cameras`)

scene.createCamera() or new Camera(scene) ? that would work for me if we have this better per-pixel API for manipulating viewport, instead of this abstract [0,1] space.


> > why expose both "point" apis and "X/Y coordinate" apis for Camera ?
> 
> Mainly for more flexible API reasons. For example when trying to convert x/y, but without need to allocate a Point instance. If anything, I think just hiding them behind `@:noCompletion` would do the trick?

Meh. You can use inline if you want to avoid Point allocation. 
API design should not be based on implementation details (apart from some very specific cases ofc).

> > just an idea, haven't thought at all the implications, but maybe we should separate Scene from CameraView
> 
> It would pretty much same as making Layers as host of the Camera. I initially went with Layers, but when it came to handling Interactives, I soon realized that having camera anywhere not on top-most level would cause a lot of problems. 

I understand, good for me.

Yanrishatum:
Quick update: Almost done, but there are 3 things left to fix yet:
* Change viewport public variables from 0...1 ratio to ones based on scene size. Most likely, internally those will remain 0...1. 
* Fix pixel bleeding. I did fix it on master, but with Camera altering how things are rendered I'll have to review the fix.
* And lastly - filters. That one in more detail.

At the moment, I got filters working on Scene by overriding `clipBounds` and fixing it to be exactly the window size (I mean, it's a scene-wide filter, so it makes sense to have entire viewport as filter bounds). As a downside - if content does not cover entire screen (just a small button in the corner, for example) - it still uses entire window as bounds. This is due to Scene not being able to tell how visual positions of objects would change based on cameras, so entire window is used to guarantee that no matter where cameras render - they won't get clipped out.
Now onto a problem of that approach. Currently it's a naive implementation, and if any transform value of the Scene instance is altered - everything falls apart and rendering breaks. Honestly, I would disallow modification of x/y/sx/sy/rotation of Scene altogether, as that function now covered by cameras. Ideas?
Next, after tinkering a bit with individual camera-wide filters - I conclude that it's better to put that feature aside for some time, as it requires preplanning and most likely will cause rewrite of how filters are handled to a degree, as cameras are not Objects. Since proper cameras are a new feature, it's reasonable to not have full feature set from the get go.
In short: `Scene.filter` - works as expected unless Scene transform is altered, `Camera.filter` - implement at a later date as an improvement of 2d camera system.

ncannasse:
Up ? ;-)

Yanrishatum:
Haven't fixed pixel bleeding yet, and for me it's the main blocker.
Also this is still unresolved @ncannasse :
> Now onto a problem of that approach. Currently it's a naive implementation, and if any transform value of the Scene instance is altered - everything falls apart and rendering breaks. Honestly, I would disallow modification of x/y/sx/sy/rotation of Scene altogether, as that function now covered by cameras. Ideas?

Yanrishatum:
So, I'd say it's done. What left unattended:
* Filters on cameras are not present. And most likely won't be for quite some time. On top of that - using filters on main scene and modifying scene transform matrix is a bad idea, and since nobody provided me with ideas on how to mitigate it - I don't care. Don't touch scene transform, use cameras. ¯\\\_(ツ)_/¯
* When doing scaleMode right/bottom/center alignments when scene size is not (integer) divisible by 2 (e.g. 799, 361, etc) - it can result in pixel bleeding. I tried to fix it, but gave up. Can be fixed later, but unless people use unconventional internal scene sizes - edge case should not resurface. If anything, can be fixed at a later date. 
* Anchor position still uses 0..1 range, because I would prefer centering camera by setting `setAnchor(0.5, 0.5)` instead of `setAnchor(s2d.width/2, s2d.height/2)`.

As a side note, would be happy if that could be merged before upcoming Ludum Dare. (But given the scale of PR - I think it requires a lot of testing, so if anything I just can use custom merge branch for LD)

ncannasse:
More changes requests/remarks. 

* We need scene transforms to work, or else to will break backward compatibility with a lot of games. Could be that the Scene transform is taken into account in Camera calculus when applySceneTransform=true, and it's set to true for default Camera

* filters on scene should work as well for same reasons. haven't tested it. If difficult to implement, one option is to keep the default current behavior with there is no camera set (camera = null)

* the fact that ctx.front2back is not taken into account in drawFilters is more like a bug. drawContent should not have a "bool" extra flag and use ctx.front2back. Also, it's drawContent (not Contents)

* regarding camera add/remove, we should keep the Object semantics : Camera should have an optional ?scene as constructor parameter, and call addCamera directly if set. It should also have a remove() function to remove itself from scene and set to null. Also remove Scene.createCamera.

* I don't think there's need to pass anchorX/Y as Camera constructor parameters as there can be set later. Also, documentation is unclear "anchor position inside viewport boundaries used for anchoring and resize compensation." --> "anchor used for anchoring" is self-referential

* XXXtoGlobal/globalToXXX should be "XXXtoScene/sceneToXXX" : I understand that it's done to match globalToLocal but since the camera has its own space it's not clear what's "global" here.

* documentation : "Internally stored as 0..1 percentile and automatically adapts on scene resize." should only be "automatically scales on scene resize." , no need to document the way it's done.

* rename Scene.setScale to setViewportScale

Let's not give up, it's not far from being merged ;-)

Yanrishatum:
> We need scene transforms to work

I'll see what I can do.

> filters on scene should work as well for same reasons

Right now: filters on Scene - ok, filters on camera - no.

> ctx.front2back in filters
> Object semantics on camera
> anchor in constructor
> global->scene
> docs
> renames

Got it.

> Let's not give up, it's not far from being merged ;-)

I've invested WAY too much effort into making decent camera system in h2d to give up at this point. My escapade with 2D cameras began pretty much same time I started to tinker with Heaps ;)

ncannasse:
Is everything ready for merge ? (according to my comments last week)

Yanrishatum:
Not yet. I've got Scene transform working with filters. With one concession: rotating/scaling the Scene will also rotate/scale where cameras are rendered in terms of viewport. It won't be noticeable with default camera, since it maps 1-to-1 onto scene viewport, but it won't be the case when camera is not mapped onto viewport entirely: 
image
Yellow is scene viewport, red is 10px border outside viewport, blue is the rest.

I still have some checks to do with this implementation, and also test how it fares with different `scaleMode` values. 

Summarize current changes:
* Now scene transforms work fine with filters, as long as it's used in legacy mode, avoid transforms (just changing x/y is fine, however) with camera usage, however.
* Filters on scene work. Filters on cameras do not (they are not implemented at all)
* `drawContent` renamed and uses `ctx.front2back` regardless if it's filters or not.
* Again changed camera instancing. Now akin to Object. It's important to note: Some methods/properties will throw an error if they are accessed without being added to Scene, because they rely on Scene instance.
* Removed anchorX/anchorY from contrusctor and changed wording on their docs.
* Renamed `global` to `scene` for clarity.
* Updated docs.

To reiterate - I still need to test the `scaleMode` to ensure everything works as I expect it to when filters are active.

ncannasse:
Tell me when finished so I can final review it ;)

Yanrishatum:
Tested it with filters + scene transform + camera movement, and it seem to work as expected. 
Hence should be ready for review.

ncannasse:
Took me some time to review all of this again. Thank you for your work !

PR #701, opened 2019-10-27
Console autocomplete

jefvel:
Added a bool `autoComplete` to Console. It will enable showing suggestions when typing commands.

Pressing `tab` will autofill the console input with the suggested command, and `return` will run the command directly:
consoleexample


jefvel:
I was wondering what the propagation stuff does, and why it's enabled by default when `globalEvents` are turned on? 
I added the propagation toggle so I can enable `globalEvents` on, and still be able to press <kbd>Tab</kbd> without the game losing focus.

PR #702, opened 2019-10-28
Audio fixes part 2: JS

Yanrishatum:
* Added initialization for `sampleOffset` to avoid `getPlayedSampleCount` returning NaN.
* Implemented support for master volume on JS and made volume control immediate instead of applied on next buffer.
* Fixed crash with stb_ogg_sound resampling due to incorrect sample count calculation (allows to play ogg files that aren't 48k).

### DIFF
--- a/hxd/snd/Manager.hx
+++ b/hxd/snd/Manager.hx
@@ -744,7 +744,7 @@ class Manager {
 		}
 
 		if (!checkTargetFormat(data, grp.mono)) {
-			size = samples * targetChannels * Data.formatBytes(targetFormat);
+			size = Math.ceil(samples * (targetRate / data.samplingRate)) * targetChannels * Data.formatBytes(targetFormat);
 			var resampleBytes = getResampleBytes(size);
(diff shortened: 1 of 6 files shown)


ncannasse:
If the MasterVolume is only implemented in JS and HL, this should be changed in openal.Driver.hasFeature accordingly

ncannasse:
Also, I prefer variables not to be public and use @:privateAccess when they need to be accessed in "internal implementation" classes, in order to separate the actual API from implementation details.

Yanrishatum:
> If the MasterVolume is only implemented in JS and HL, this should be changed in openal.Driver.hasFeature accordingly

It actually always reported that it supported master volume in Emulator, even while it's clearly false.

> Also, I prefer variables not to be public and use @:privateAccess when they need to be accessed in "internal implementation" classes, in order to separate the actual API from implementation details.

`ctx`, `destination` and `masterGain` are public for ability to set up custom output chain in WebAudio. Same goes for `gain` node. But I can make them private if you want (since it's going to become legacy fallback solution with new driver anyway)

PR #706, opened 2019-10-28
Added slider handle to slider bounds

jefvel:
fixes https://github.com/HeapsIO/heaps/issues/705

### DIFF
--- a/h2d/Slider.hx
+++ b/h2d/Slider.hx
@@ -15,7 +15,6 @@ class Slider extends h2d.Interactive {
 		tile.dy = (height - 4) >> 1;
 
 		cursorTile = h2d.Tile.fromColor(0xCCCCCC, 5, height);
-		cursorTile.dx = -2;
 	}
 
 	function set_minValue(v) {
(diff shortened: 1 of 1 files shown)


ncannasse:
A more correct fix would be to take into account the cursor X position, as done in Render:

var dx = Math.round( (value - minValue) * (width - cursorTile.width) / (maxValue - minValue) );

and use dx instead of cursorTile.dx

jefvel:
I did some changes to the slider, so that the cursor will drag the handle without making it jump if you click on the edge of it. It makes it easier to do precise changes to the slider value. 

Example:

before:

slider_old

after:

slider_new

ncannasse:
Please fix indentation : heaps code should use tabs

jefvel:
It's done

ncannasse:
Please don't change current implementation by forcing the tile dx.  Some custom implementations might prefer to have their own placement for slide cursor, or might have the same cursor tile shared between several sliders.

jefvel:
Alright! It now returns the handle dx to normal after drawing

Yanrishatum:
Regarding tile DX offsetting just for one draw: I wonder why `RenderContext.drawTile` won't just accept `offsetX:Float = 0, offsetY:Float = 0` as an extra params? That would simplify cases such as this, because you'd just pass dx as `emitTile(ctx, cursorTile, dx, 0)`. This is pretty reasonable improvement for singular tile drawcall. I actually had a need for such thing several times already, and only solution was either do as Slider does - offset dx/dy, then revert or clone Tile outright.

h2d/Slider.hx
```diff
@@ -38,33 +37,53 @@ class Slider extends h2d.Interactive {
 		super.getBoundsRec(relativeTo, out, forSize);
 		if( forSize ) addBounds(relativeTo, out, 0, 0, width, height);
 		if( tile != null ) addBounds(relativeTo, out, tile.dx, tile.dy, tile.width, tile.height);
+		if( cursorTile != null ) addBounds(relativeTo, out, getDx(), cursorTile.dy, cursorTile.width, cursorTile.height);
```
ncannasse: should be cursorTile.dx + getDx() no ?

PR #710, opened 2019-10-31
WIP: Basis support

lbergman:
Support for loading basis files (https://github.com/BinomialLLC/basis_universal).

It is working as is, but still a WIP for a few reasons:

1. Support is tested for DXT, ATSC and PVRTC, but having issues with ETC.
2. Integrate with asset system. 
3. Include basis transcoder and wasm binary.

...

1.
Not sure when ETC is needed, but will see if I can find why it is not working. DXT, ATSC and PVRTC seems to cover desktop and mobile browsers well though.

2.
Since we handle loading without depending on assets system in heaps, this is not really a priority for us and seems a bit tricky to do. One reason being that basis adds another async stage to the process (first the usual loading, and then running wasm), so rather than adding another supported format all asset handling needs to be modified to be async both when loading and parsing. Any help on this would be much appreciated.
At the moment is has to be used through BasisTextureLoader.getTexture. Current solution we use in our games looks like this:
```
function getImageTexture(path:String, ext:String, name:String):js.lib.Promise<h3d.mat.Texture> {
	return switch ext {
		case 'basis':
			final bytes = _fileSystem.get(path).getBytes().getData();
			hxd.res.BasisTextureLoader.getTexture(bytes);
		default:
			final bytes = _fileSystem.get(path).getBytes();
			js.lib.Promise.resolve(hxd.res.Any.fromBytes(name, bytes).toTexture());
	}
}
```

3.
Unsure if .js and .wasm needed should be checked in to heaps and added to build folder as needed.

### DIFF
--- a/h3d/impl/GlDriver.hx
+++ b/h3d/impl/GlDriver.hx
@@ -799,7 +799,7 @@ class GlDriver extends Driver {
 		case SRGB, SRGB_ALPHA: hasFeature(SRGBTextures);
 		case R8, RG8, RGB8, R16F, RG16F, RGB16F, R32F, RG32F, RGB32F, RG11B10UF, RGB10A2: #if js glES >= 3 #else true #end;
 		case S3TC(n): n <= maxCompressedTexturesSupport;
-		case ASTC(_), ETC(_), S3TC(_), PVRTC(_): #if js true #else false #end;
+		case ASTC(_), ETC(_), PVRTC(_): #if js true #else false #end;
 		default: false;
 		}
 	}
--- a/h3d/impl/GlDriver.hx
+++ b/h3d/impl/GlDriver.hx
@@ -127,6 +127,7 @@ class GlDriver extends Driver {
 	static var UID = 0;
 	public var gl : GL2;
 	public static var ALLOW_WEBGL2 = true;
+	public var textureSupport:hxd.PixelFormat;
 	#end
 
 	#if (hlsdl||usegl)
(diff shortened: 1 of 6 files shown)


lbergman:
ETC is working now and added parsing of bytes to get image width/height from basis bytes.

However that presented another obstacle with regards to integrating with heaps asset system. You can encode several images with different size into one basis file. However hxd.res.Image is built for resources with a single image.

So considering both that and the async parsing of images, hxd.res support seems like a big invasive task, so I probably skip that. Since it is so different than other formats and only applicable to js target, at least for our needs it is fine to access it via BasisTextureLoader.

That leaves one point, and it is eventual inclusion of js and wasm for transpiling.

Let me know if you have any ideas about that or the MR i general. If it is of interest merge in I will add relevant documentation and example.

lbergman:
We are using this in our latest games, and despite being a very limited implementation it is at least working well to minimize filesize on the web, getting similar size to jpg but with alpha support.

However there is one problem that is really limiting the usefulness. With PVRTC it seems like source png's need to have pre multiplied alpha. However it seems like heaps is not automatically correcting the blending in case of PMA, see https://github.com/HeapsIO/heaps/issues/726

What we currently do is correcting by redrawing texture with a shader applied when loading, but this means that the texture is no longer compressed in memory, defeating some of the purpose with using basis in the first place.

A couple of other minor issues:
* ETC1 doesn't support alpha channel, and to have proper support for it one should blend an rgb and alpha texture. Considering it only seems to be needed on some very old Android devices it doesn't seem worth the hassle. 
* capturePixels in GlDriver fails. Does this work with S3TC in HL? I get mismatch with data length and size.
* Have not tested with mipmaps, probably will not work.

lbergman:
Closing, new implementation for ktx2 support https://github.com/HeapsIO/heaps/pull/1254

PR #711, opened 2019-11-02
Audio fixes part 3: WebAudio driver

Yanrishatum:
* Implements new default audio driver for JS target that works directly with WebAudio
* Can be disabled with `-D useal` flag, then legacy AL emulation is used.
* Changed audio context unlock to not eat up user input.
* Both AL emulator and WebAudio context use shared `hxd.snd.webaudio.Context` as audio provider to avoid code duplication.
* Implemented experimental pitch, low-pass and spatialization effects. (See notes below)
* Disabled resampling to native frequency for WebAudio driver, since it supports automatic resampling under the hood.

#### Effects
* LowPass won't produce same sound as with HL/AL, but at least it's working.
* Spatialization uses built-in Panner/Listener combo, meaning that velocity is not supported.
* Pitch is a tough nut, because of Absolutely Sane And Not Idiotic WebAudio Standard. Changing playbackRate mid-playback causes a lot of synchronization issues. I've managed to make it more or less usable, but position reporting does bug out for a frame after changing the pitch (and when switching buffer) and it does occasionally create gaps/overlaps between buffers. I tried my best without hacks utilizing AudioWorklets (which require separate JS file, because web workers). Main issue is inability to know current position of the buffer from WebAudio, and in the end relying on manual calculation which can and is often does not align with reality. At least it's an a-rate parameter, and not k-rate. I wish there was a buffer queue node that could deal with synchronization.

### DIFF
--- a/hxd/impl/TypedArray.hx
+++ b/hxd/impl/TypedArray.hx
@@ -5,6 +5,7 @@ package hxd.impl;
 #if haxe4
 typedef Float32Array = js.lib.Float32Array;
 typedef Uint16Array = js.lib.Uint16Array;
+typedef Int16Array = js.lib.Int16Array;
 typedef Uint8Array = js.lib.Uint8Array;
 typedef ArrayBuffer = js.lib.ArrayBuffer;
 typedef Uint32Array = js.lib.Uint32Array;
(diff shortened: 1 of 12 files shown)


ncannasse:
Thanks !

PR #714, opened 2019-11-07
Make Drawable.color to not affect the filter

### LINKED ISSUE #670: Outline shader/filter uses wrong color on h2d.Text
.filter = new Outline(5, color);

There is some weird behavior on the h2d.Text. The outlines should use the same color.

image

Also changing the textcolor afects the applied color.

image

image

Yanrishatum:
* `Drawable.color` affects only object itself and not the filter.
* `Object.alpha` affects transparency of both object and filter.
* `Drawable.color.a` can be used to manipulate object transparency while inside filter.

Change probably should be tested internally in case it borks stuff in Northgard or something ;)

Fixes #670

### DIFF
--- a/h2d/Drawable.hx
+++ b/h2d/Drawable.hx
@@ -8,7 +8,6 @@ class Drawable extends Object {
 
 	/**
 		The color multiplier for the object. Can be used to adjust individually each of the four channels R,G,B,A (default [1,1,1,1])
-		Alpha channel can be used to manipulate object transparency but leave transparency of active filter intact.
 	**/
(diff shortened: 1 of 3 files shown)


ncannasse:
I would not change documentation at all here ;)

Yanrishatum:
Why? I personally prefer detailed documentation of nuanced things like that. ;)

ncannasse:
Matter of taste. I think the API is quite clean. This should be explained
in filter maybe.

Le ven. 8 nov. 2019 à 17:26, Pavel Alexandrov <notifications@github.com> a
écrit :

PR #721, opened 2019-11-25
Fixed window restore size after fullscreen toggle

jefvel:
Fixes https://github.com/HeapsIO/heaps/issues/717


hxd/Window.hl.hx
```diff
@@ -339,6 +357,7 @@ class Window {
 			46 => K.QWERTY_PERIOD,
 			47 => K.QWERTY_SLASH,
 			59 => K.QWERTY_SEMICOLON,
+			60 => K.QWERTY_BACKSLASH,
```
jefvel: This was unrelated to the issue, but fixes that you can detect the key above tab on Macs when using the english layout. 
image

ncannasse:
I have mixed feelings about this. I think we need somehow to review our fullscreen management instead.

@Yanrishatum any insight ?

Yanrishatum:
Honestly I feel like this should be handled by hlsdl/hldx internally rather than being something Heaps should bother with. And by looking at hlsdl, I see that it tries to do exactly that, but for some reason fails to restore window size: https://github.com/HaxeFoundation/hashlink/blob/f96bdb3a5b45956505b9e12724d989b5f270d9c6/libs/sdl/sdl.c#L459-L464

As for revising - I can't say there's much to be revised apart from adding ability to set specific mode instead of doing on/off switch. Exclusive + BorderlessFS + Windowed is already standard choice for display modes. I can say however that window management in general needs some touchups. At least exposing of damn `title` property.

jefvel:
Ah of course. That's a much better place to handle it. I fixed it in this merge request: https://github.com/HaxeFoundation/hashlink/pull/332

I could look into adding support for changing window titles of windows and the possibility to switch between exclusive and borderless windowed mode too.

PR #724, opened 2019-11-28
Fixed AlphaMap shader (incorrect channel)

deepnight:
For some reason, alphaMap used Blue channel instead of Alpha as source 🤔

### DIFF
--- a/h3d/shader/AlphaMap.hx
+++ b/h3d/shader/AlphaMap.hx
@@ -8,14 +8,19 @@ class AlphaMap extends hxsl.Shader {
 		@param var texture : Sampler2D;
 		@param var uvScale : Vec2;
 		@param var uvDelta : Vec2;
+		@const var useAlphaChannel : Bool;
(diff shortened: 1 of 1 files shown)


ncannasse:
The alphaMap is not always a map with alpha, it can be an RGB map with gray values representing the amount of alpha, hence the confusion. I think it would require a boolean const to differentiate between the two.

deepnight:
Is it ok to have an IF inside the shader fragment() ?

```
if( useAlphaChannel ) 
  pixelColor.a *= texture.get(calculatedUV * uvScale + uvDelta).a;
else 
  pixelColor.a *= texture.get(calculatedUV * uvScale + uvDelta).b;
```

Yanrishatum:
In that case it should be marked as const: `@const var useAlphaChannel:Bool`. In that case HXSL will optimize out the if and will generate unique shader for each variation.

deepnight:
I added the change to this PR

PR #727, opened 2019-12-01
Default letterSpacing to 0

### LINKED ISSUE #720: [RFC] Default Text.letterSpacing to 0
Currently it's at 1px, and was justified by legacy font format, e.g. Littera fonts, which do not provide advance values. Actually decent format (BMF) provides advance of the glyph, removing the need in that offset and actually said offset starts to create problems, as it creates spacing where there should be none. As a result working with newer and preferred font formats (especially after fontgen tool implementation) it is a hindrance, as that value have to be adjusted for every single text instance.
Proposal:
* Set `Text.letterSpacing` to 0.
* To preserve spacing for Littera fonts - add 1 pixel to advance during parsing of glyphs.
* Because of that, bfnt cache in .tmp should be deleted for littera fonts to receive proper pixel advances.

Now more free-worded background reasoning.
Every single goddamn time I have to either keep in mind that spacing is 1 by default and reset it to 0 each time I create text instances or make `Util.text` method that wraps this operation. That is annoying as hell and was annoying ever since implementation of adequate BMF format support. That inferior garbage format should not create issues for people who use intended ways to generate fonts. I asked Threthaller if that variable is relied on being 1 by default, and it looks like no, apart from providing spacing for littera fonts.

Yanrishatum:
* Defaults `Text.letterSpacing` to 0
* Invalidates all bitmap fonts (since we can't detect which one is affected)
* Revert weird hotfix from 2014.
* Added `HtmlText.imageSpacing` (defaults to 1)

There was one particular part I missed when pitching the change. HtmlText uses letterSpacing to pad `` tiles. Hence - introduced `HtmlText.imageSpacing` which remains defaulted to 1.

Closes #720

### DIFF
--- a/h2d/HtmlText.hx
+++ b/h2d/HtmlText.hx
@@ -5,6 +5,10 @@ import h2d.Text;
 class HtmlText extends Text {
 
 	public var condenseWhite(default,set) : Bool = true;
+	/**
+		Spacing after <img> tags in pixels.
+	**/
+	public var imageSpacing(default,set):Float = 1;
 
 	var elements : Array<Object> = [];
 	var xPos : Float;
@@ -121,7 +125,7 @@ class HtmlText extends Text {
 				newLine = true;
 			case "img":
 				var i = loadImage(e.get("src"));
-				len = (i == null ? 8 : i.width) + letterSpacing;
+				len = (i == null ? 8 : i.width) + imageSpacing;
 				newLine = false;
 			case "font":
 				for( a in e.attributes() ) {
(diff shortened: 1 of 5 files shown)


ncannasse:
@trethaller ok for merge on my side once DB/NG are ready

trethaller:
I checked and both look fine. Some conflicts in HtmlText which I resolved like this https://github.com/trethaller/heaps/commit/65380d8b8da4cdbe108f090aeee059beaa91d016

Issue #542, opened 2019-02-14
Cannot set window title


samw3:
I can't see a way to set the title of the window via Heaps when using hlsdl.

Yanrishatum:
Currently title can be set only at compile time with a `-D windowTitle=mytitle` define. (From public API. With access to internals it is possible to access SDL Window instance and set title manually at runtime)

ncannasse:
Maybe that's something @RobertBorghese would be interested contributing (at least on the hlsdl/hldx level)

RobertBorghese:
Huh, surprised I didn't throw this in sooner. Should be a piece of cake and would be nice to have. 

I'll add it on the HashLink-side of things and get back to you in a bit. 👌

trethaller:
For reference, as Pavel said it's currently possible using privateAccess, this will work for SDL and DX:

```@:privateAccess hxd.Window.getInstance().window.title = title;```

ncannasse:
Close by #732

PR #732, opened 2019-12-04
added displayMode and title property to window. deprecated setFullscreen

jefvel:
closes https://github.com/HeapsIO/heaps/issues/730

### DIFF
--- a/hxd/Window.hl.hx
+++ b/hxd/Window.hl.hx
@@ -9,6 +9,13 @@ import hxd.Key in K;
 typedef DisplayMode = sdl.Window.DisplayMode;
 #elseif hldx
 typedef DisplayMode = dx.Window.DisplayMode;
+#else
+enum DisplayMode {
+	Windowed;
+	Borderless;
+	Fullscreen;
+	FullscreenResize;
+}
 #end
 
 //@:coreApi
--- a/hxd/Window.hl.hx
+++ b/hxd/Window.hl.hx
@@ -410,10 +410,16 @@ class Window {
 	}
 
 	function get_title() : String {
+		#if (hldx || hlsdl)
 		return window.title;
+		#end
+		return "";
 	}
 	function set_title( t : String ) : String {
+		#if (hldx || hlsdl)
 		return window.title = t;
+		#end
+		return "";
 	}
 
 	static var inst : Window = null;
--- a/hxd/Window.hl.hx
+++ b/hxd/Window.hl.hx
@@ -5,11 +5,11 @@ import hxd.Key in K;
 #error "You shouldn't use both -lib hlsdl and -lib hldx"
 #end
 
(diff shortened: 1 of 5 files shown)


hxd/Window.hl.hx
```diff
@@ -5,6 +5,12 @@ import hxd.Key in K;
 #error "You shouldn't use both -lib hlsdl and -lib hldx"
 #end
 
+enum DisplayMode {
+	Windowed;
+	Borderless;
+	Fullscreen;
+}
+
```
Yanrishatum: Better replace with `typedef DisplayMode = sdl/dx.Window.DisplayMode` to avoid unnecessary conversion between the two.

hxd/Window.hl.hx
```diff
@@ -380,6 +394,34 @@ class Window {
 
 	#end
 
+	function get_displayMode() : DisplayMode {
+		var mode = window.displayMode;
+
+		if ( mode == windowedMode) return Windowed;
+		if ( mode == borderlessMode) return Borderless;
+		if ( mode == fullScreenMode) return Fullscreen;
```
Yanrishatum: Should be gated with `#if (hldx || hlsdl)`, since `window` does not exists with `usesys`.

jefvel:
> Additional notes:
> I'm uncertain if JS should also implement `title`.

Not sure either, but I thought it could have its uses. If the game runs embedded in a website as an iframe it shouldn't change the title anyway.

hxd/Window.hl.hx
```diff
@@ -5,6 +5,12 @@ import hxd.Key in K;
 #error "You shouldn't use both -lib hlsdl and -lib hldx"
 #end
 
+#if hlsdl
+typedef DisplayMode = sdl.Window.DisplayMode;
+#elseif hldx
+typedef DisplayMode = dx.Window.DisplayMode;
+#end
```
jefvel: Oh yeah, this will fail on usesys (`DisplayMode` undefined). Should it just be an enum on that target?

Yanrishatum:
This also resolves #542, as it fullfills public API for title change.

hxd/Window.hl.hx
```diff
@@ -380,6 +395,27 @@ class Window {
...
+
+	function set_displayMode( m : DisplayMode ) : DisplayMode {
+		#if (hldx || hlsdl)
+		window.displayMode = m;
+		#end
+		return displayMode;
+	}
+
+	function get_title() : String {
+		return window.title;
```
ncannasse: I think we need to restrict get/set of title to hlsdl || hldx

ncannasse:
Thanks, just one minor thing and we're good for merge ;)

jefvel:
Ah yes, missed that one! It's fixed now :+1:

PR #734, opened 2019-12-08
Implement async audio decoder

Yanrishatum:
* Based off `target.threaded` compiler flag enables async buffer decoder.
* Can be forced to use non-threaded mode with `-D snd_sync` flag.
* Implementation does not affect short sounds, they are still decoded in main thread. 
* I'm not sure about dispose code and in general how clean the implementation is, as I'm not very adept with multithreaded programming.

### DIFF
--- a/hxd/snd/Manager.hx
+++ b/hxd/snd/Manager.hx
@@ -20,6 +20,12 @@ class Source {
 	public var streamBuffer : haxe.io.Bytes;
 	public var streamStart : Int;
 	public var streamPos : Int;
+	#if (!snd_sync && target.threaded)
+	// 0 = Not decoded OR was decoded and processed by main thread
+	// 1 = Decode in progress
+	// 2 = Decoding is complete by decoder thread but wasn't processed by main thread.
+	public var streamDecoded : Int;
+	#end
 
 	public function new(driver : Driver) {
 		id      = ID++;
@@ -96,6 +102,11 @@ class Manager {
 	var effectGC          : Array<Effect>;
 	var hasMasterVolume   : Bool;
 
+	#if (!snd_sync && target.threaded)
+	var decoderThreadInst:sys.thread.Thread;
+	// var decoderQueue
+	#end
+
 	public var suspended : Bool = false;
 
 	private function new() {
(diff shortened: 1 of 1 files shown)


ncannasse:
TBH I'm not yet sure exactly what I want to do with that.

I think your code might have some issue when we request decoding from some part of the stream, but then the stream goes silent (virtual) and then kicks back later with a tag=2 (decoded) while we want to decode another part of it.

ncannasse:
Also I think it should also support "small sounds" decoding since that's where most of the sounds (and then CPU) come from.

ncannasse:
Adding @trethaller to the discussion

Yanrishatum:
I'm not sure about implementation with small sounds. Yet now that you mentioned it I agree that offloading decoding of small sounds to separate thread is better solution (reminds me of a Synth Riders which loads effect SFX in main thread mid-track, causing entire game to freeze. Which is extremely bad, especially given that it's a VR game).
It also creates a delay (of at least 1 frame) when playing the sound for first time or after it was GCd.

Yanrishatum:
Implementation of async decode for short SFX is worth doing, but there should be a way to manually mitigate the caveat of 1+ frame delay with uncached SFX. Something like `SFXCache` akin to `ModelCache`? That also would enable preloading of all required SFX on a scene beforehand (except streaming ones, as they're handled separately).

ncannasse:
One way would be to trigger caching as soon as its played, but then you will cache even if it's virtualized. But maybe that's better this way so it will be ready in cache in case it become unvirtualized later.

Yanrishatum:
After taking a quick look, there are no easy solutions for small sounds to be processed on separate thread. First of all, to avoid race condition, access to buffer cache should be gated with `Lock`. Then there is more glaring issue - buffers are GCd as soon as their `refs = 0` (In case cache is full), and until Manager queues the sound, `refs` would be 0, hence it should prepare buffer instance before sending it off to decoder with refcounter increased as to avoid it being cached as soon as decoder done with it but main thread is yet to try to queue it, and it should be done correctly, otherwise there is a risk of memory leak with buffers. 
Right now I think it's outside of the scope of this PR (I aimed specifically at streaming audio decoding). I can open separate issue about async SFX decoder. Since Heaps should be able to work with both threaded and non-threaded targets, trying to implement it without prior planning would be a bad idea. :) 
Streamed decoder is easy to implement mainly because there is no risk of race condition, and it's very straightforward change as there was already delayed decode implemented, and on top of that it's delay-tolerant as next chunk gets decoded before current one ends. That is not the case with short sounds (and with first chunk of streamed audio btw).

PR #748, opened 2020-01-11
add curveThru

nanjizal:
curveThru useful when creating quadratic curves that go through a list of points, often more intuitive api than curveTo and user friendly.

____________________________________________________________________________________________________________
( Note: I would suggest that curveTo be renamed to quadTo, and cubicCurveTo be renamed to curveTo to match js & java naming rather than flash, but have gone with your implemented naming scheme. )

( side note:
If your interested in an implementation for svg eclipseArc then I can pull my implementation, it is unfortunately more complex than your current drawEllipse implementation, but would enable path implementations.
https://github.com/nanjizal/justPath/blob/master/justPath/EllipseArc.hx // last method not required.
used here
https://github.com/nanjizal/justPath/blob/master/justPath/SvgPath.hx#L309
tested here
https://github.com/nanjizal/TrilateralBazaar/blob/master/screenshots/ellipticalArcExamples.png )

### DIFF
--- a/h2d/Graphics.hx
+++ b/h2d/Graphics.hx
@@ -565,6 +565,22 @@ class Graphics extends Drawable {
 	}
 
 	/**
+          * Similar to curveTo drawing a quadratic Bezier but allows bx, by to be a point on the curve.
+	 * @param bx mid point X on curve
+	 * @param by mid point Y on curve
+	 * @param cx end X
+	 * @param cy end Y          
+         */ 
+	public function curveThru( bx: Float, by: Float, cx: Float, cy: Float ): Void {
+		var ax = tmpPoints.length == 0 ? 0 :tmpPoints[ tmpPoints.length - 1 ].x;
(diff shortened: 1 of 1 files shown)


ncannasse:
This is very non standard way to deal with curves, you can instead have it into your own library. (with `using`)

nanjizal:
Yep I have just added the algorithm geom.curve.Bezier.quadraticThru as you suggest.
<https://github.com/nanjizal/geom/blob/master/src/geom/curve/Bezier.hx#L84>
A generic version that does not suppose toolkit, segmentation technique or dimensions.

I already have a version in trilateral2 that needs better segmentation, but yet to work out how to setup triangle x,y,z buffers in Heaps or which Prim that would be best to rework for my needs, ideally I wanted to pass a Float32Array positions and Int32Array for colors, graphics is too high level for my interest. Suggested approach or tip would be appreciated.

PR #749, opened 2020-01-12
Updated h2d.col.Line with angle() and length() methods

Syopic:
Line.angle()  - The angle between a line and the x-axis
Line.length()  - Distance between line points

### DIFF
--- a/h2d/Graphics.hx
+++ b/h2d/Graphics.hx
@@ -569,8 +569,8 @@ class Graphics extends Drawable {
 	 * IvanK Lib port ( http://lib.ivank.net )
 	 */
 	public function curveTo( bx : Float, by : Float, cx : Float, cy : Float) {
-		var ax = tmpPoints[ tmpPoints.length - 1 ].x;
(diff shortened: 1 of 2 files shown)


ncannasse:
I prefer if we don't store dx/dy, because the points can be mutated, and also because it consumes more memory if you have many lines only to save a few fast subtract.

Syopic:
Yes, really. I will delete the cache. 
Are these methods necessary at all? I was surprised that there was no method for calculating the absolute length of the line. By analogy I can add these methods to h2d.col.Segment.

ncannasse:
Yes I think they can be added, we have not yet designed the whole col API with all methods, only the ones we need when we need them. Please also avoid reformatting existing code in the PR.

ncannasse:
typo to `length`

ncannasse:
Also use hxd.Math.distance

Syopic:
More writings than benefitі of this pool request, but thank you)

Issue #606, opened 2019-04-30
h2d.Flow overflow


trethaller:
I suggest to modify the `overflow` property on h2d.Flow. It is only useful in combination with `maxWidth` or `maxHeight`. 
- `Expand`: This is the current, default mode. Children larger than `maxWidth`/`maxHeight` will expand the flow size.
- `Limit`: Limits the bounds reported by the flow using `maxWidth` or `maxHeight`, if set. This is means children larger than max size will draw outside of their parent bounds.
- `Hidden`: Like `limit`, but using a h2d.Mask to prevent children from drawing outside
- `Scroll`: See below

Not sure yet how this would behave with constraints coming from parents rather than current flow.

Any thoughts @ncannasse @Fl0xer @D0pu ?

ncannasse:
We already have an overflow:bool so both expand/limit are already available. We can change it to an enum to add hidden I don't see any problem with it.

Note: in order to enable scrollable content, we should have a way to make the difference between the "virtual" size (including hidden/limited parts) and the "display" one. ATM we only have calculatedWidth/Height used for things such as innerWidth/outerHeight

trethaller:
Missed that we already had one 🤦‍♂️. Good news is that it already does what I asked! So we're just missing `Hidden` and `Scroll`. Edited the original to reflect the enum names, rather than the anticipated domkit CSS ones

ncannasse:
Do we really need hidden/scroll to be handled as part of Flow ? (which is quite complex already)
I think we should instead use some parent component. For instance Mask.

Yanrishatum:
I actually agree that it's probably not the place for having overflow, as separate class would suit better. On top of that, if separate - it can be use for other than Flow.
In fact! I already have some basic classes that do exactly that in Heeps:
https://github.com/Yanrishatum/heeps/blob/master/h2d/ui/ScrollArea.hx
https://github.com/Yanrishatum/heeps/blob/master/h2d/ui/ScrollText.hx

ncannasse:
I think this should somewhat be merged with h2d.Mask, so the inner content can be scrolled.

Yanrishatum:
I can port implementation of ScrollArea to Mask (it's a bit hacky precisely because I had to copy-paste some of Mask code in order to make it work), but there is a question of what features have to be ported.
* Should is support scrollBounds? E.g. custom limiter on how far it can be scrolled as to prevent overscroll.
[...]

PR #751, opened 2020-01-14
Mask.maskWith API / h2d.Flow.overflow enum

Yanrishatum:
Ref: #606 
* Implemented `Mask.maskWith` and `Mask.unmask` methods. 
Had to use static stack to hold renderZone state, otherwise using that API would break mask-in-mask (and most likely filters). Alternatively I would say it's possible to change `setRenderZone` to `pushRenderZone(x, y, w, h, inherit)` stack, allowing for more robust render-zone handling. (With inherit = true, new renderZone would be clipped by existing renderZone)
* Changed `h2d.Flow.overflow` to an enum
* Implemented Hidden overflow.
* Added showcase of advanced masking to Mask sample.

I'm not touching `FlowOverflow.Scroll`, too much have to be accounted, I'll leave it to the person who actually assigned to the task. ;)

### DIFF
--- a/h2d/Mask.hx
+++ b/h2d/Mask.hx
@@ -7,6 +7,10 @@ class Mask extends Object {
 	static var renderZoneStack:Array<RenderZoneStack> = [];
 	static var renderZoneCaret:Int = 0;
 
+	/**
+		Masks render zone based off object position and given dimensions.
+		Should call `Mask.unmask()` afterwards.
+	**/
 	@:access(h2d.RenderContext)
 	public static function maskWith( ctx : RenderContext, object : Object, width : Int, height : Int, scrollX : Float = 0, scrollY : Float = 0) {
 
@@ -48,6 +52,9 @@ class Mask extends Object {
 		ctx.setRenderZone(x1, y1, x2-x1, y2-y1);
 	}
 
+	/**
+		Unmasks prviously masked area from `Mask.maskWith`.
+	**/
 	public static function unmask( ctx : RenderContext ) {
 		if (renderZoneCaret == 0) throw "Too many unmask()";
 		var inf = renderZoneStack[--renderZoneCaret];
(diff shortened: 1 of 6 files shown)


kLabz:
Any news on this?

h2d/Mask.hx
```diff
@@ -95,7 +159,7 @@ class Mask extends Object {
 				c.posChanged = true;
 			posChanged = false;
 		}
-		addBounds(relativeTo, out, 0, 0, width, height);
+		addBounds(relativeTo, out, scrollX, scrollY, width, height);
```
ncannasse, line 162: This does not seem correct.

h2d/Mask.hx
```diff
@@ -1,7 +1,71 @@
 package h2d;
 
+private typedef RenderZoneStack = { hasRZ:Bool, x:Float, y:Float, w:Float, h:Float };
+
 class Mask extends Object {
 
+	static var renderZoneStack:Array<RenderZoneStack> = [];
+	static var renderZoneCaret:Int = 0;
```
ncannasse, line 8: These should be moved to ctx

ncannasse:
See my two comments ;)

h2d/Mask.hx
Yanrishatum: Should I remake `RenderContext.setRenderZone` to `pushRenderZone` then?

ncannasse:
Sure.

h2d/Mask.hx
Yanrishatum: Why? With that, Mask will report proper content bounds. Sample below showcases it. Green outline is mask boundaries, red is reported boundaries.
Currently: 
image
With `-scrollX, -scrollY` it does not clip bounds properly and actually offsets them:
image
And with `0, 0` it just reports offset boundaries of Mask:
image
So everything is correct. :)

ncannasse:
Thanks!

PR #757, opened 2020-01-27
Add lookat to quat

jefvel:
it creates a quaternion rotated to a direction


ncannasse:
How does it compare with initLookAt ? Also, please do not include openURL in the PR

jefvel:
It's the same as initLookAt, but will point in the Z axis. So it's not really needed. I could achieve my same goal with initMoveTo in the end on my part.

PR #760, opened 2020-01-30
Fix incorrect pop filter on offscreen object

Togimaro:
While working on DKH we encountered a bug.
If an object with a filter is offscreen the RenderContext keep pushing the filter without ever poping it, resulting in a weird behavior and a leak.


ncannasse:
Thanks!

PR #763, opened 2020-02-06
feat: adding extended support for instancing also in webgl1 devices s…

lusius:
Since instanced rendering is only supported trough ANGLE-extension there needs to be a check for this.

ANGLE_instanced_arrays is supported by 98% of mobile devices and is directly compatible with the core version for webgl2. 

This update uses this and makes sure devices that supports instancing gets it even if they are running webgl1 context.

### DIFF
--- a/h3d/impl/Driver.hx
+++ b/h3d/impl/Driver.hx
@@ -93,6 +93,10 @@ enum Feature {
 		Supports rendering in wireframe mode.
 	*/
 	Wireframe;
+	/*
+		Supports instanced rendering
+	*/
+	InstancedRendering;
 }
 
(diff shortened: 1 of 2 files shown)


ncannasse:
I'm planning at some point to use Haxe 4 WebGL 2 definition, so could you change the PR not to use dynamic functions ? You can try using Reflect.setField instead. Thanks!

lusius:
Sure can do.

lusius:
Ping @ncannasse

ncannasse:
Sorry for the delay. Can you replace `final` by `var` so it works on older haxe versions?

PR #768, opened 2020-02-24
Fixed JS version of hxd.File.browse on Chrome

deepnight:
For some obscure reason, the method used to upload a file to an app on Chrome didn't work sometimes (the Open window would pop-up, but the onChange callback wasn't called).

The new version adds the <input> element to the actual <body> and it stays there, hidden.

### DIFF
--- a/hxd/File.hx
+++ b/hxd/File.hx
@@ -162,6 +162,7 @@ class File {
 					}
 				}
 				onSelect(b);
+				input.remove();
 			}
 			input.click();
 		#else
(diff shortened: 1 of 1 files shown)


ncannasse:
Could we have instead a modified version that removes it once the callback is triggered ? I think that keeping an element in the tree this way was causing some other problems. Maybe also try by simply keeping the element in a static Array (for Garbage Collector) and removing before callback triggered.

deepnight:
I removed the input when callback is called. It seems to work fine, but it needs a little bit more public testing. I will see if I don't get any report on my app for the next few days.

deepnight:
It's working so far for me. It could be merged, if you're ok too :)

ncannasse:
Thanks!

PR #769, opened 2020-03-10
add Circle primitive

tong:
Example: http://tmp.disktree.net/heaps-primitives/  
Use `W` to toggle wireframe.

### DIFF
--- a/h3d/prim/Circle.hx
+++ b/h3d/prim/Circle.hx
@@ -1,0 +1,35 @@
+package h3d.prim;
+
+import h3d.col.Point;
+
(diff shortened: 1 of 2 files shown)


ncannasse:
Maybe it should be Disc instead of Circle ?

tong:
Not sure.
As a blender user i would expect Circle, but:

Blender: Circle
Maya: Disc
Threejs: Circle

ncannasse:
I think Disc is more appropriate.

ncannasse:
Thanks!

PR #772, opened 2020-03-21
Only display stacktraces in a window on Windows platform

jstolarek:
Consider the following incorrect program using heaps:

```haxe
class Main extends hxd.App {
  static var b : Array<Int>;

  static public function main():Void {
    //b = new Array<Int>();
    b.push(42);
  }
}
```

That program tries to add element to an array that wasn't allocated, which is obviously incorrect. Compiling and running under Linux causes the program to hang in an infinite loop:

```
$ haxe -main Main -lib heaps -cp . -hl foo.hl
$ hl foo.hl
Null access
Called from $Main.main (Main.hx line 6)
```

The problem is in the [System.reportError](https://github.com/HeapsIO/heaps/blob/2b3efb27a55dd17901d3be47b146011d6bf763d8/hxd/System.hl.hx#L155) function that attempts to display a stacktrace in a window. However, UI subsystem is not implemented in hashlink on Linux ([see here](https://github.com/HaxeFoundation/hashlink/blob/master/libs/ui/ui_stub.c)). In particular, heaps expects that a [UI loop will eventually return a `Quit`](https://github.com/HeapsIO/heaps/blob/2b3efb27a55dd17901d3be47b146011d6bf763d8/hxd/System.hl.hx#L183):

```
while( hl.UI.loop(true) != Quit )
    timeoutTick();
```

but that will never be the case on Linux, since the [Linux UI stub always returns `HandleMessage`](https://github.com/HaxeFoundation/hashlink/blob/master/libs/ui/ui_stub.c#L39) (see [here](https://github.com/HaxeFoundation/haxe/blob/4.0.5/std/hl/UI.hx#L123) for definition of `LoopResult`).

```
HL_PRIM int HL_NAME(ui_loop)( bool blocking ) {
  return 1;
}
```

**Solution**: only create a UI window on Windows platform. With this patch running `hl foo.hl` on Linux produces a stacktrace on the console and terminates normally:

```
$ hl foo.hl
Uncaught exception: Null access
Called from $Main.main(Main.hx:14)
Called from fun$3997(?:1)
```

Behaviour on Windows should remain unchanged.

### DIFF
--- a/hxd/System.hl.hx
+++ b/hxd/System.hl.hx
@@ -158,10 +158,11 @@ class System {
 		#if usesys
 		haxe.System.reportError(err + stack);
 		#else
+		Sys.println(err + stack);
+
 		if ( Sys.systemName() != 'Windows' )
 			return;
 
-		Sys.println(err + stack);
 		if( dismissErrors )
 			return;
 
--- a/hxd/System.hl.hx
+++ b/hxd/System.hl.hx
@@ -158,7 +158,7 @@ class System {
 		#if usesys
 		haxe.System.reportError(err + stack);
 		#else
-		Sys.println(err + stack);
+		Sys.stderr().writeString(err + stack + "\n");
 
 		if ( Sys.systemName() != 'Windows' )
 			return;
--- a/hxd/System.hl.hx
+++ b/hxd/System.hl.hx
@@ -158,6 +158,9 @@ class System {
 		#if usesys
 		haxe.System.reportError(err + stack);
 		#else
+		if ( Sys.systemName() != 'Windows' )
+			return;
+
 		Sys.println(err + stack);
 		if( dismissErrors )
 			return;


hxd/System.hl.hx
```diff
@@ -158,6 +158,9 @@ class System {
 		#if usesys
 		haxe.System.reportError(err + stack);
 		#else
+		if ( Sys.systemName() != 'Windows' )
+			return;
```
Yanrishatum: Return should be after printing of the error to the system console, not before.

jstolarek: That was my initial thought but then the error information is printed twice:

```
hl freeze_bug.hl
Null access
Called from $Main.main (Main.hx line 14)
Uncaught exception: Null access
Called from $Main.main(Main.hx:14)
Called from fun$3997(?:1
```

jstolarek: And BTW the error should be printed on stderr, not stdout, but that's orthogonal to the subject of this PR.

ncannasse:
That seems weird that it is printed twice as this should entirely override error handling. I want to investigate this before merging. Tell me if you find the reason.

ncannasse:
Ah, I think I get it. It depends if the error is on app initialization (in which case it will get printed twice) or in main loop (in which case it will print before continuing the loop, but not abort the app with Uncaught exception).

jstolarek:
But this error is raised in the `main` function. Doesn't that count as the main loop? Or do you mean `hxd.App.update()` as the main loop?

jstolarek:
Ok, I figured out what's going on. There are two problems. Firstly, the reason why my original program reports two errors is because `reportError` prints the contents of the exception stack but it does not empty that stack. Therefore, when the execution ends hashlink detects that the exception stack is not empty and reports an exception for the second time, marking the error message as an uncaught exception. I think this is in fact correct - the exception has indeed not been caught in the program. And indeed it is necessary to print the error in `reportError` before aborting the function - otherwise caught exceptions will not be printed at all. I made the necessary change and also changed the printing to go to the stderr rather than stdout - these changes are now pushed to this PR and ready to be merged. Let me know if you want me to rebase & squash.

I have also noticed a second error printing problem. Consider the following program:

```haxe
class Main extends hxd.App {
  static var b : Array<Int>;

  override function init() {
      b.push(42);
  }

  static public function main():Void {
    new Main();
  }
}
```

Running it produces two identical error messages:

```
$ hl freeze_bug.hl
heaps was updated
Null access
Called from Main.init (Main.hx line 5)
Called from hxd.App.~setup.2 (hxd/App.hx line 142)
Called from hxd.App.loadAssets (hxd/App.hx line 170)
Called from hxd.App.setup (hxd/App.hx line 140)
Called from h3d.Engine.onCreate (h3d/Engine.hx line 246)
Called from h3d.impl.GlDriver.~init.0 (h3d/impl/GlDriver.hx line 1526)
Called from haxe.$Timer.~delay.0 (/dane/sandboxes/haxe/std/haxe/Timer.hx line 143)
Called from haxe.$Timer.~__constructor__.0 (/dane/sandboxes/haxe/std/haxe/Timer.hx line 76)
Called from haxe.$MainLoop.tick (/dane/sandboxes/haxe/std/haxe/MainLoop.hx line 174)
Called from hxd.$System.runMainLoop (hxd/System.hl.hx line 134)
Called from haxe.$Timer.~delay.0 (/dane/sandboxes/haxe/std/haxe/Timer.hx line 143)
Called from haxe.$Timer.~__constructor__.0 (/dane/sandboxes/haxe/std/haxe/Timer.hx line 76)
Called from haxe.$MainLoop.tick (/dane/sandboxes/haxe/std/haxe/MainLoop.hx line 174)
Called from haxe.$EntryPoint.processEvents (/dane/sandboxes/haxe/std/haxe/EntryPoint.hx line 99)
Called from haxe.$EntryPoint.run (/dane/sandboxes/haxe/std/haxe/EntryPoint.hx line 125)
Null access
Called from Main.init (Main.hx line 5)
Called from hxd.App.~setup.2 (hxd/App.hx line 142)
Called from hxd.App.loadAssets (hxd/App.hx line 170)
Called from hxd.App.setup (hxd/App.hx line 140)
Called from h3d.Engine.onCreate (h3d/Engine.hx line 246)
Called from h3d.impl.GlDriver.~init.0 (h3d/impl/GlDriver.hx line 1526)
Called from haxe.$Timer.~delay.0 (/dane/sandboxes/haxe/std/haxe/Timer.hx line 143)
Called from haxe.$Timer.~__constructor__.0 (/dane/sandboxes/haxe/std/haxe/Timer.hx line 76)
Called from haxe.$MainLoop.tick (/dane/sandboxes/haxe/std/haxe/MainLoop.hx line 174)
Called from hxd.$System.runMainLoop (hxd/System.hl.hx line 134)
Called from haxe.$Timer.~delay.0 (/dane/sandboxes/haxe/std/haxe/Timer.hx line 143)
Called from haxe.$Timer.~__constructor__.0 (/dane/sandboxes/haxe/std/haxe/Timer.hx line 76)
Called from haxe.$MainLoop.tick (/dane/sandboxes/haxe/std/haxe/MainLoop.hx line 174)
Called from haxe.$EntryPoint.processEvents (/dane/sandboxes/haxe/std/haxe/EntryPoint.hx line 99)
Called from haxe.$EntryPoint.run (/dane/sandboxes/haxe/std/haxe/EntryPoint.hx line 125)
```

The reason is the way main loop is implemented in `System.runMainLoop()`:

```haxe
while( true ) {
    try {
        hl.Api.setErrorHandler(reportError); // set exception trap
        @:privateAccess haxe.MainLoop.tick();
        if( !mainLoop() ) break;
    } catch( e : Dynamic ) {
        hl.Api.setErrorHandler(null);
        reportError(e);
    }
    ...
}
```

When an exception is raised it is reported twice: first by a call to `reportError` set as the error handler, and then second time by an explicit call to `reportError` in the `catch` block. If you empty the `catch` block then the error is correctly reported once. Alternatively, you could give up on calls to `hl.Api.setErrorHandler` and explicitly call `reportError` in the `catch` block, which also gives a desired result. I admit I don't entirely follow all the details here. In particular, if I empty the `catch` block it is still necessary to have a call to `hl.Api.setErrorHandler(reportError);` inside the loop. If you place the call to `setErrorHandler` before the `while` loop then no error printing happens, so it seems like `reportError` is not called at all. I don't understand why that is the case.

jstolarek:
Is there anything that I could/should do to push this forwards?

jstolarek:
Thanks for merging!

I wonder about the problem outlined in my previous comment, i.e. double printing of a stack trace. The fix should be a oneliner, I'm just not sure which of two possible solutions to pick.

Issue #743, opened 2020-01-06
[RFC] Library-provided default converts


Yanrishatum:
After introduction of FileConverter, only way to add custom converts is via props.json, while in old system it could be done via macro. This change essentially disabled ability of external libraries (read: Heeps) to add it's own default converts (e.g. gif convert) without extreme hackery of patching FileConverter class. 
Proposal:
Expose API to add that allows adding custom Converts via macro api (or at compilation time). While converts are registered, they cannot be enabled without mentioned manual involvement.

Possible solutions:
* Alter register method with extra parameter: `defaultConvert:Bool = false`, which would then affect if said convert is enabled by default. Then, instead of hard-coded list FileConvert grabs a list of converts that are enabled by default.
* Use same approach as `hxd.res.Config` and expose that list for macro to edit. This is more or less old approach, but provides less clarity on how it supposed to work.

Example use case:
I have a GifConvert class in Heeps that converts gif files into png atlases accompanying with giff metadata file that holds information about frame delays, image dimensions, etc. That is done to ease up loading of gif animations, as I use them extensively in my workflow when working with pixelart (asking artists to "just throw me that gif" is far more comfortable for them compared to doing sprite sheets, as pretty much always PA does not utilize semi-transparency) and using convert to preprocess them into format that more optimal for Heaps to render is logical. But right now it's not possible to enable said convert by default unless I use props.json and specify it directly.
(Also same convert makes me want to make ability to store more than one file into .tmp with one Convert, as right now I have to store png file in main res directory)

ncannasse:
Any converter that is compiled with the "program" (either the game or a hide plugin or a commandline PAK builder) can be used using props.json without requiring any change to FileConverter.

As "default converts configuration", I think you're referring to https://github.com/HeapsIO/heaps/blob/master/hxd/fs/FileConverter.hx#L43

One possibility would be to have this default configuration stored in a JSON that would be embedded at compile time, and allow a macro to bind another file instead of the std one. Would that work for you ?

ncannasse:
There was some discussion regarding storing metadata into .tmp , I think @D0pu even had a working implementation at some point but I don't think it was merged

Yanrishatum:
>> As "default converts configuration", I think you're referring to

[...]

PR #777, opened 2020-03-30
Add ability to enable converts without props.json

Yanrishatum:
Basic implementation of adding converts without need to introduce props.json to the project. 
Introduces `hxd.fs.FileConverter.addConfig` which takes structure compatible with props.json and doc with sample of usage.
In my case, I use it for gif -> spritesheet + frame data conversion:
```haxe
// in GifImage resource class, since registering those in Convert class requires manually
// causing that class to resolve statics at some point.

  static var _ = hxd.fs.Convert.register(new hxd.heeps.GifConvert());
  static var __ = hxd.fs.FileConverter.addConfig({
    "fs.convert": {
      "gif": "giff"
    }
  });
```
Ref: #743 (close if satisfied with such implementation, or leave if can be improved later)

### DIFF
--- a/hxd/fs/FileConverter.hx
+++ b/hxd/fs/FileConverter.hx
@@ -34,17 +34,46 @@ class FileConverter {
 	var defaultConfig : ConvertConfig;
 	var cache : Map<String,Array<{ out : String, time : Int, hash : String, ver : Null<Int> }>>;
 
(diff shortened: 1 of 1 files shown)


Yanrishatum:
Pinging due to upcoming Ludum Dare. Also, what happened to travis checks?

ncannasse:
Thanks!

PR #779, opened 2020-04-03
feat: Making separate alpha handling configurable in h2d.RenderContext

lusius:
We had issues with inconsistent setting of alpha blending in our game between 3d and 2d and being able to set alphahandling separately in Alpha channel solved it for us.

### DIFF
--- a/h2d/RenderContext.hx
+++ b/h2d/RenderContext.hx
@@ -17,6 +17,9 @@ class RenderContext extends h3d.impl.RenderContext {
 	public var onLeaveFilter : h2d.Object->Void;
 
 	public var tmpBounds = new h2d.col.Bounds();
+
+	public static var separateAlphaSrc = [BlendMode.Alpha => h3d.mat.Data.Blend.One, BlendMode.Add => h3d.mat.Data.Blend.One];
+
 	var texture : h3d.mat.Texture;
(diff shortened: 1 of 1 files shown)


ncannasse:
I'm curious which problem exactly you got and how you solved it. Alpha handling with filters is quite tricky indeed but I _think_ I got it right and would surely want to know if some cases are not covered.

lusius:
Hm it's quite tricky to answer this question because I've always
experienced that the gl.blendFunc(gl.SRC_ALPHA, gl.ONE_MINUS_SRC_ALPHA)
always gives me the correct blending when comparing to photoshop. The issue
with the current standard setting for 2d scene is when you have a render to
texture in a completely empty buffer (no color, no alpha). And then you
render a texture with "glass". It showed up as darker because it's assuming
background is black and also loosing the alpha data.

So short answer, in my experience with directX and webgl "gl.SRC_ALPHA,
gl.ONE_MINUS_SRC_ALPHA" is the method that gives you the most common wanted
result and therefore i wanted to be able to set that as default.

thank you again for your amazing work.

>

ncannasse:
Once the texture is rendered and needs to be layered on top of another one, you need to use (one, one minus src alpha) since the color has already been premultiplied with it's alpha. That's what we do when merging a texture that's been filtered.

PR #792, opened 2020-04-16
BDF font parser. Extends Font.

projectitis:
Created new type h2d.BDFFont that extends h2d.Font.  Use it to load/parse a BDF font files, then use the BDFont as you'd normally use Font.

Example usage:

`var f = new h2d.BDFFont( hxd.Res.My_BDF_font_file.entry.getBytes() );`
`var tf = new h2d.Text(f);`

A BDF font file (created by Adobe, possibly in the 1980s) is a self-contained text file that contains packed bit data (with optional smoothing/alpha) and does not have a separate PNG bitmap.  I tried (hard) to add BDF parsing to hxd.fmt.bfnt.FontParser but because of the reliance that FontParser has on external bitmap/PNG files, I was not able to get it to work that way. However, in the future it would be really nice if someone smarter than me could combine BDFFont with Font and integrate with the Resource.toFont method.

Instead I created a new Font type. It parses the BDF file and creates it's own glyph tile. It uses a basic but pretty effective packing algorithm to generate (almost) the smallest tile it can to save space. 

FontForge is one example program that is able to generate BDF files. The benefit of BDF files is that the entire font is self-contained. A negative of BDF is that it does not have support for kerning.

### DIFF
--- a/h2d/BDFFont.hx
+++ b/h2d/BDFFont.hx
@@ -1,0 +1,438 @@
+package h2d;
+
+import hxd.BitmapData;
+import haxe.Int32;
+import h2d.Font;
+
+/**
+ * Intermediate representation of a glyph. Only used while
+ * parsing a BDF font file.
+ */
+class BDFFontChar{
+    public var code : Int;
+    public var x : Int;
+    public var y : Int;
+    public var width : Int;
+    public var height : Int;
+    public var xoffset : Int;
+    public var yoffset : Int;
+    public var stride : Int;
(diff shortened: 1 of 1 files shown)


Yanrishatum:
Since bitmap font Resource is separate from Font instance, you can just declare your own Resource that takes BDF files and have same `toFont()` method that would return regular Font instance. No need for wonky Font instance extension.

projectitis:
Good idea. Much better solution. Will close this request and investigate that instead.

Issue #602, opened 2019-04-19
Graphical Corruption on Mobile (iOS)


thejustinwalsh:
Experiencing graphical corruption on an iPhone Xs when rendering 3D objects in both the World Sample and the Skin sample, and potentially others however I only tested those two thus far.

I've attached screen recordings below.

World iOS native target:
![world-app-ios](https://user-images.githubusercontent.com/72912/56428891-d671f900-6275-11e9-89ab-9973f0922e7b.gif)

World JS target:
![world-web-ios mov](https://user-images.githubusercontent.com/72912/56428901-deca3400-6275-11e9-8f86-0b7c72f63e77.gif)

thejustinwalsh:
I have merged in #599 and am experiencing the same results as above vs head.

ncannasse:
We made some fixes wrt GpuParticles recently.

Could you confirm that:
- disabling Gpu Particles fixes the problem
- recent changes in GpuParticles fixes the problem

ncannasse:
Closing due to lack of answer. Assuming fixed.

NewKrok:
Hey, I just checked it with my phone (iPhone 6s) - https://heaps.io/samples/world.html - And it's the same. I was not sure how old is this build there so I built it from 1.7.0 and it's the same there.
http://flashplusplus.net/haxe/world3d/index.html

In my own project it's the same too and when I dont change the camera's y position than it looks it works.
Can it be a problem with the projectedPosition calculation in the BaseMesh shader? I have weird issues on desktop browser too and when I set the z manually than it helps but it's impossible to maintain (I will try to investigate it but my knowledge is limited in this topic).

![image](https://user-images.githubusercontent.com/13141660/70867926-4bb7da00-1f7b-11ea-8e2a-35fcb0f42bda.png)

ncannasse:
This might be related to z buffer precision given we are using a very far z
plane. Does changing s3d.camera.zFar work better?

NewKrok:
The default settings in the sample is:
`s3d.camera.zNear = 1;`
`s3d.camera.zFar = 100;`

I modified the demo, you can check it too here:
http://flashplusplus.net/haxe/world3d-zfar/index.html
(When I modify the zFar/zNear and I'm using a cameracontroller it has no any affect - maybe it was my fault - so I just added a dynamic cam position)

For me it looks like more or less the same
[...]

PR #797, opened 2020-04-21
Enable high shader precision in WebGL

lbergman:
Shaders are currently compiled using mediump which is fine in many cases, but will result in precision issues on many devices. For example shadows becomes very strange if not using highp, and issues like #602 is probably related.

I added MAX_PRECISION constant to GlDriver so that one can choose if highp should enabled or not. I set it to mediump by default to ensure that behavior is same as before. I think highp would be a sensible default but don't want to risk causing issues for existing projects.

So to enable high precision, somewhere before compiling shaders one has to do `GlDriver.MAX_PRECISION = 'highp';`to enable.

And to check highp support, for example to see if it is a good idea to enable shadows:
``` 
public function hasHighPrecision() {
	final driver:h3d.impl.GlDriver = cast h3d.Engine.getCurrent().driver;
	return driver.getMaxPrecision() == 'highp';
}
```


ncannasse:
The PR is fine, however I'm curious whereas we should allow such high level global change from mediump to highp instead for instance allowing @highp as variable qualifier - that would only be enabled if the devices supports it ofc.

What's your take on this?

lbergman:
Do you mean that it should be possible to set per shader?

Agreed, I was thinking the same, but figured it was a bit more complicated change, and that regardless it makes sense to be also be able to have a default max precision if supported.

For our games I didn't notice any obvious performance issue with switching to highp, and for example three.js uses 'highp' per default as well. But I will do more testing on more devices and if we find it necessary to control precision per shader I will make another MR.

ncannasse:
I meant more per-variable.
But if several popular libraries are using highp by default when supported
I think we should do the same. So please update your PR with "highp" as the
default max precision.
Also, we might want set highp for GL desktop as well?

lbergman:
Yes, it would be nice to expose variable precision as well. 
I got a report about degraded performance on one machine in a build including this change. Could be something else, but I like to investigate to see if we need more finegrained control over precision right now, in which case I will look in to adding that.

But perhaps a separate PR in that case? Should I just modify this to use highp default and apply to all gl targets for now?

ncannasse:
> I got a report about degraded performance on one machine in a build including this change.

Let's wait for more feedback on this first

lbergman:
Finally got feedback from the the machine that started dropping frames, and it seems like when  reverting to mediump performance was back to normal.

Now this is a somewhat problematic machine that often shows worse performance than other machines we test with, so our conclusion is that highp is still a sensible default for shaders, but that we would like to be able to have control per shader and variable.

Very busy with some releases right now, but will have a look at if I can get a PR together as soon as I get a chance.

zommerfelds:
Hi!

I'm running into issues drawing shapes (h2d.Graphics) and sprites with pretty simple scenes. Positions are off and shapes distorted.

Similar description to this issue in different engine: https://forum.defold.com/t/terrible-distortions-just-2-screens-away-from-origin-android-def-3946/52325

It sounds like setting `highp` would fix the issue. Is there a plan to support this with a global option?

lbergman:
Now defaulting to highp for vertex shaders since #960, and this is no longer needed.

PR #800, opened 2020-04-21
Add Polygons example: show examples of all poly in h3d.prim

neokeld:
In this PR, i add an example of polygons from h3d.prim : 

- Grid
- Cube
- Cylinder
- Disc
- Geosphere
- Sphere

I also try to add arrow key movements but i'm not sure of the implementation.
Feel free to criticize.

### DIFF
--- a/samples/Polygons.hx
+++ b/samples/Polygons.hx
@@ -1,12 +1,43 @@
-import h3d.prim.Grid;
-import h3d.prim.GeoSphere;
-import h3d.prim.Disc;
-import h3d.prim.Sphere;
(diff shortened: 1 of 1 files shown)


ncannasse:
Don't use addEventTarget but instead override update for keys and add a CameraController (see Base3d).

neokeld:
Thanks a lot ! I've corrected this using overriding as you suggested and this has greatly simplified the implementation.

ncannasse:
No need for castle inspector, it's been deprecated ;)

neokeld:
Thx i've removed it :)

PR #807, opened 2020-04-23
Added a throw in BinaryLoader for unsupported platforms

deepnight:
Btw, there's no way to do a simply HTTP async url request in Haxe? 🤔


ncannasse:
There's haxe.Http but system platforms does not always provide a "main event loop" for asynchronous calls

PR #814, opened 2020-05-02
add Mat2 type

### LINKED ISSUE #810: HXSL Mat2 type is not supported
Im trying to create a 2D rotation matix. The GLSL code is
```
mat2 rotate2d(float _angle){
    return mat2(cos(_angle),-sin(_angle),
                sin(_angle),cos(_angle));
}
```
In HXSL this would be
```
function rotate2d( angle:Float ):Mat2 {
	return 	mat2(cos(angle),-sin(angle),
                	sin(angle),cos(angle)); 
}
```
I get the error message
```Unsupported type Mat2```

It would be cool to be able to work with this type and it shouldn't be too hard to implement. (I hope)

0xA3:
Resolves #810

There will be conflicts with [my other pull request](https://github.com/HeapsIO/heaps/pull/812) because some code changes are at similar places. But should be easy to resolve.

### DIFF
--- a/hxsl/AgalOut.hx
+++ b/hxsl/AgalOut.hx
@@ -827,6 +827,7 @@ class AgalOut {
 	function regSize( t : Type ) {
 		return switch( t ) {
 		case TInt, TFloat, TVec(_), TBytes(_), TBool: 1;
+		case TMat2: throw "Mat2 is not supported in AGAL";
 		case TMat3, TMat3x4: 3;
 		case TMat4: 4;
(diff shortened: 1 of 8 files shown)


ncannasse:
- same remark as your other PR regarding   mat x  vec , please remove corresponding code
- please add Mat2 as the end of types. this way it won't break existing serialized shaders
Thanks!

0xA3:
Ok thanks.
I made the changes. I also found out that AGAL doesn't support Mat2. But I think it nevertheless makes sense to have Mat2 for GL and DX.
The constructor with 4 floats has been testet for GL and DX.

ncannasse:
Thanks!

PR #816, opened 2020-05-07
Add Save.savePath

Yanrishatum:
* Allows to simplify saving data on PC.
* Ensures that save directory does exist as to not cause exception when saving to `saves/savename` where `saves` does not exist.
* Adds `-D savePath=path/to/save/dir` flag support for hl (except usesys)

Ideally, I'd want to also add binding to automatically use `%appdata%` (and linux/mac alternatives), but that would require implementing [SDL_GetPrefPath](https://wiki.libsdl.org/SDL_GetPrefPath) for hlsdl and also defining of org/app for it to work. For now it at least would allow to specify base path instead of passing path in the save name.
My case: Settings during development, which I want to store in appdata, since build going to be stored in shared folder, hence saving data in cwd would cause a lot of pain with version conflicts and such.

### DIFF
--- a/hxd/Save.hx
+++ b/hxd/Save.hx
@@ -6,6 +6,12 @@ package hxd;
 **/
 class Save {
 
+	/**
+		Save path location for sys targets. Defaults to current working directory.  
+		Use `-D savePath=path/to/save/directory` to set this variable on Heaps boot.
+	**/
+	public static var savePath:String = "";
+
 	static var cur = new Map<String,String>();
 	#if flash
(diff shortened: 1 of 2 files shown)


ncannasse:
I don't like it much, hxd.Save has already been made so the methods can be overidden (they are dynamic) so you should be able to write this quite particular change without having to change original hxd.Save.

I would consider instead a static enum Location that would be like:
- `Cwd` (default)
- `UserData` (os specific)
- `Custom( dir : String )`

PR #817, opened 2020-05-08
Add h2d.RenderContext.clipRenderZone

projectitis:
This is a method to push a new render zone to the h2d RenderContext, but it first clips (intersection) to the current render zone so that the current zone is still respected. I use this for rendering child objects in a UI hierarchy so that children cannot render outside their parent (all the way up the tree).


projectitis:
Fixed formatting to match by other PR (BDF fonts). Squashed/rebased.

PR #818, opened 2020-05-10
fix: handle normalized SDL touch coordinates

qkdreyer:
https://wiki.libsdl.org/SDL_TouchFingerEvent

N.B. : Since we're rounding the normalized value using 100 factor, we can have an 1% windowWidth pixel loss (`792px` vs `799px` for `800px` default windowWidth), maybe we should increase the factor by 10 there and here, what do you think @ncannasse ?

https://github.com/HaxeFoundation/hashlink/blob/a30209ef5020fce4b3b1cda3e7a4fed2edd7f118/libs/sdl/sdl.c#L186-L203

### DIFF
--- a/hxd/Window.hl.hx
+++ b/hxd/Window.hl.hx
@@ -255,12 +255,24 @@ class Window {
 			else
 				((c & 0x0F) << 18) | (((e.keyCode >> 8) & 0x7F) << 12) | (((e.keyCode >> 16) & 0x7F) << 6) | ((e.keyCode >> 24) & 0x7F);
 		case TouchDown if (hxd.System.getValue(IsTouch)):
+			#if hlsdl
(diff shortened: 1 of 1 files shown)


ncannasse:
I agree it should be scaled up by 100 , from *100 to *10000

qkdreyer:
https://github.com/HaxeFoundation/hashlink/pull/379

ncannasse:
Could you do  `static inline var TOUCH_SCALE = #if (hl_ver >= version("1.12.0")) 10000 #else 100 #end`
Would require compiling with -D hl-ver=1.12.0 , but at least that doesn't break current HL.

ncannasse:
#if hsdl around it please ? :D

qkdreyer:
Done ;)

ncannasse:
Thanks!

Issue #819, opened 2020-05-12
TextInput : support for standard shortcuts


Fl0xer:
- Ctrl + A : select all
- Ctrl + arrow : navigate word by word
- Enter : skip a line when allowed and when it will be handled
- Escape : remove focus from the text input (for now, it's Enter)

kLabz:
I started working on this before seeing this issue. For CTRL+A there is #823. I'll see for ENTER and ESC.

As for CTRL+Arrow, I have something locally that navigates per word but is limited to "real" spaces (handles multiple spaces). What rules should we follow for defining a word? Should we separate by:
* One or more space (implemented locally)
* Any whitespace character
* Any non-alphanumeric character
* Any non-[alphanumeric character, including `_` and accents (and some other ascii symbols?)]

ncannasse:
Spaces as defined by hxd.CharSet seems good enough.

kLabz:
Got it, will do that tonight.

PR #823, opened 2020-05-18
[h2d] TextInput: select all text on CTRL+A

kLabz:


kLabz:
Missing `selectionSize = 0;`, and I will open a PR with all shortcuts from #819 instead

PR #824, opened 2020-05-19
[HLSDL] Avoid conflicts between keycodes

### LINKED ISSUE #713: TextInput under SDL
Here are a couple weird behaviours when using TextInput with SDL (using an AZERTY french keyboard) :
- Typing a $ or ¤ do a Home (cursor goes back to position 0) each time.
- Typing a £ delete everything else but the other £ characters.

Everything works fine under DirectX.

kLabz:
Currently, sdl keycodes are translated as-is to "windows virtual-key codes", which are not a 1-1 map and cause conflicts (#713, for example).

This PR avoids the automatic mapping of the 2048 first keycodes for hlsdl, and maps manually the missing control keys (backspace, tab, etc.). The actual characters are still available via `e.charCode`.

Fixes #713

### DIFF
--- a/hxd/Window.hl.hx
+++ b/hxd/Window.hl.hx
@@ -45,7 +45,7 @@ class Window {
 	var curMouseX = 0;
 	var curMouseY = 0;
 
-	static var CODEMAP = [for( i in 0...2048 ) i];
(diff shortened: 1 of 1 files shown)


ncannasse:
The fix will break custom handling of some special keys that are not covered by "standard" key codes, such as ² on french keyboard, etc.

PR #825, opened 2020-05-19
TextInput shortcuts

kLabz:
This PR implements most points from #819, except multiline support which is more involved. Maybe a separate issue should be open for multiline TextInput?


ncannasse:
Thanks !

ncannasse:
And yes a separate issue is necessary for multiline input

PR #826, opened 2020-05-25
Fix Blender exports

Yanrishatum:
* Altered string parsing on binary FBX. Now it cuts off anything past `0` byte if it's present in a string. This is due to Blender inserting `\0\1AnimCurveNode` extra data to curve node names, clearly expecting it being cut away on parsing.
* Removed Blender downscale hack: It's no longer required, as it's possible to export with proper unit scale (see wiki notes).

### DIFF
--- a/hxd/fmt/fbx/BaseLibrary.hx
+++ b/hxd/fmt/fbx/BaseLibrary.hx
@@ -234,8 +234,6 @@ class BaseLibrary {
 			case "LastSaved|ApplicationName": app = p.props[4].toString();
 			default:
 			}
-		if( app.indexOf("Blender") >= 0 && unitScale == 1 && originScale == 1 )
-			scaleFactor *= 0.01; // Blender in meters exports FBX to centimeter
 
 		if( scaleFactor == 1 && geometryScaleFactor == 1 )
 			return;
--- a/hxd/fmt/fbx/BaseLibrary.hx
+++ b/hxd/fmt/fbx/BaseLibrary.hx
@@ -234,10 +234,8 @@ class BaseLibrary {
 			case "LastSaved|ApplicationName": app = p.props[4].toString();
 			default:
 			}
-		#if fbx_legacy_blender
 		if( app.indexOf("Blender") >= 0 && unitScale == 1 && originScale == 1 )
 			scaleFactor *= 0.01; // Blender in meters exports FBX to centimeter
-		#end
 
 		if( scaleFactor == 1 && geometryScaleFactor == 1 )
 			return;
--- a/hxd/fmt/fbx/BaseLibrary.hx
+++ b/hxd/fmt/fbx/BaseLibrary.hx
@@ -234,8 +234,8 @@ class BaseLibrary {
 			case "LastSaved|ApplicationName": app = p.props[4].toString();
 			default:
 			}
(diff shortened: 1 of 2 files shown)


ncannasse:
What about legacy FBX ? we don't want to break Northgard / Darksburg models because they were not reexported :-/

Also, ZERO_BYTE might not work on some platforms as they don't allow a \0 in the middle of the String.

Yanrishatum:
@ncannasse Would gating it behind `blender_legacy` flag suffice? Because if left as is, properly exported FBX would be downscaled, while they shouldn't be.

ncannasse:
Is there no way to differentiate these two fbxs ? (with or without scaling applied)

Yanrishatum:
Just to be sure, I'd need to know with what settings legacy FBX were exported, to ensure it would reliably detect them. It may be possible to detect, but not guaranteed. I'll remove the gate for now, as to fix animation exports, and will see if I can get a reliable detection of legacy and new fbx.

trethaller:
Here are the settings we have used so far
image

neokeld:
I made a similar change as @Yanrishatum locallly some time ago to try to resolve https://github.com/HeapsIO/heaps/issues/790 and https://github.com/HeapsIO/heaps/issues/789 and it resolves this two issues (which means Unity doesn't apply the 0.01 scaling when it imports Blender created files).
It also means that after the change HeapsIO will be able to import any FBX from Blender with this settings without problem : 
- Apply Scalings FBX All
- Apply Scalings Custom Scale
- Apply Scalings Unit Scale
- Apply Scalings All Local

To reply to the problem identified par @ncannasse I've created a PR on @Yanrishatum branch : https://github.com/Yanrishatum/heaps/pull/3
I use the ApplicationVersion block in FBX to check the version of Blender before applying scaling. Thus it will work with old exported FBX (with Blender) and newer (2.79 and more).

Yanrishatum:
I'm hesitant to rely on app version, since it's still possible to export invalid FBX with newer versions. (Also docs specifically mention that you should use `FBX` units).
Another difference that is present is when exporting with different scalings:
`GlobalSettings.Properties70.UnitScaleFactor` and `OriginalUnitScaleFactor`. Tested with `Scale` value set to 1.0 and 5.0:
* `All Local`: 1 in both cases
* `FBX Unit Scale`: 100 in both cases.
* `FBX Custom Scale`: Represents `Scale` value of 1 and 5 respectively.
* `FBX All`: 100 and 500.

So proper solution is most likely to rely on what `UnitScaleFactor` and `OriginalUnitScaleFactor` present.
```haxe
		if( app.indexOf("Blender") >= 0 && unitScale == originScale )
			scaleFactor = unitScale / 100;
```
That would upscale and downscale everything to expected units regardless of the selected export option. (`FBX All` / `FBX Unit Scale` is still preferred over `All Local` as later causing broken animations)

`unitScale == originScale` is most likely redundant, as Blender always seem to export both with same value. Also `scaleFactor` in general probably should be calculated as `unitScale / originScale` instead of `unitScale == 100 && originScale == 1 ? 100 : 1`.

ncannasse:
I agree we should use fbx internal data instead of relying on app version unless the version exporter had an actual bug that needs to be addressed.

ncannasse:
Thanks !

PR #830, opened 2020-06-01
Expand 2D matrix API

Yanrishatum:
* Add skewing transformation support:
`initSkew(sx, sy)`, `skew(sx, sy)`, `skewX(sx)`, `skewY(sy)`
* Add per-axis methods:
`translateX(x)`, `translteY(y)`, `prependTranslateX(x)`, `prependTranslateY(y)`, `scaleX(sx)`, `scaleY(sy)`


ncannasse:
Thanks!

PR #832, opened 2020-06-04
Add autocomplete suggestions to console args

jefvel:
Added  `?availableOptions : Array<String>`, and `?resolveOptions : String… -> Array<String>` to console arguments so you can do

```
console.addCommand("test", "cool args", [
            { 
                name : "test_arg", 
                t : AString, 
                availableOptions: [ "arg1", "cool"]
            },
            {
                name: "test_arg2",
                t: AString,
                resolveOptions: (e) -> {
                    if (e.charAt(0) == "e") {
                        return [ "epic", "eritrea" ];
                    }

                    if (e.charAt(0) == "d") {
                        return [ "donut", "dang" ];
                    }
                    
                    return null;
                }
            }
        ], (arg1, arg2) -> {
           trace(arg1, arg2);
        });
```

### DIFF
--- a/h2d/Console.hx
+++ b/h2d/Console.hx
@@ -14,6 +14,8 @@ typedef ConsoleArgDesc = {
 	name : String,
 	t : ConsoleArg,
 	?opt : Bool,
+	?availableOptions : Array<String>,
+	?resolveOptions : String -> Array<String>,
 }
 
 class Console #if !macro extends h2d.Object #end {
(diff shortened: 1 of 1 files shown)


ncannasse:
That's getting a bit complex and quite specific maybe ? Instead, using enums for auto completion would be better maybe, since with reflexion you can list all possible values.

jefvel:
Yeah, true. I can make those changes when I have the time

PR #836, opened 2020-06-16
Fix javascript error in GlDriver

rrohrer:
The original code generated `if((typeof)(WebGLDebugUtils) != "undefined")` which doesn't run in a browser. This should generate `if((typeof WebGLDebugUtils) != "undefined")` which will.

I'm a haxe novice and am not sure if this is the correct fix, but it appears to work.


rrohrer:
Hmm, it seems master build is failing for other reasons :(

ZwodahS:
I encountered this as well. Suspect the fix is `js.Syntax.typeof(untyped WebGLDebugUtils)` instead.

nanjizal:
I expect you need   #if( haxe<4.2 ) __js__ ...  #else js.Syntax...

PR #842, opened 2020-06-25
API documentation: h2d package

Yanrishatum:
This PR is done thanks to donations from 01010111, Kenny Goff, Weast, Ryan Rohrer and Daniel Sun.

I took on an effort of fixing one of long-standing issues community had with Heaps: Little to no documentation. Because I won't be able to cover everything in one sweep, it will be done in parts, and this is h2d package part. 

While current docs are a draft version and not the final one (tho I would say majority of it would be left untouched), I would welcome anyone to provide commentary on unclear moments, points out phrasing issues, grammar errors and such.

I did not touch documentation of the following classes:
* ~`h2d.Camera` - because new camera PR. :) Why would I touch that.~ Merged already and is documented.
* `h2d.CdbLevel` - it's an external dependency not many use, documented to a degree and majority of it's API is rather self-evident.
* `h2d.Layers` - because it's already documented.
* ~`h2d.KeyFrames` - as with CdbLevel it's an external dependency.~ Partially documented.
* ~`h2d.filter.Ambient` - because I couldn't figure out what exactly it supposed to do.~ Thanks to Cannasse for explanation.

# Current state of PR

* [x] Cover root `h2d` package classes
* [x] Cover `h2d.col` package
*  Cover `h2d.domkit` package~
* [x] Cover `h2d.filter` package
* [x] Go trough entire PR and add documentation to private methods that are meant to be overridden with usage of `@:dox(show)`
* [x] Go trough entire PR and redraft it. Fix typos, unclear statements and in general make it more readable.

Temporary hosting of the WIP docs can be found here: https://wb.yanrishatum.ru/heaps/api/

### DIFF
--- a/h2d/Console.hx
+++ b/h2d/Console.hx
@@ -2,23 +2,41 @@ package h2d;
 
 import hxd.Key;
 
+/**
+	Console argument type.
+**/
 enum ConsoleArg {
 	AInt;
 	AFloat;
 	AString;
 	ABool;
+	/**
+		`AEnum` type acts as `AString` but restricts allowed values to provided list of strings.
+	**/
 	AEnum( values : Array<String> );
 }
 
+@:dox(hide)
 typedef ConsoleArgDesc = {
 	name : String,
 	t : ConsoleArg,
 	?opt : Bool,
 }
 
+/**
+	`h2d.Console` provides simple debug console integration.
+
+	By default comes with 2 commands: `help` and `cls`, which print help message
+	describing all command and clear the screen respectively
+
+	To add custom commands, use `Console.add` and `Console.addCommand` methods.
+**/
 class Console #if !macro extends h2d.Object #end {
 
 	#if !macro
+	/**
+		The timeout in seconds before log will automatically hide after last message. ( default : 3 )
+	**/
 	public static var HIDE_LOG_TIMEOUT = 3.;
 
 	var width : Int;
@@ -35,9 +53,18 @@ class Console #if !macro extends h2d.Object #end {
 	var logIndex:Int;
 	var curCmd:String;
 
+	/**
+		The text character which should be pressed in order to automatically show console input.
+	**/
 	public var shortKeyChar : Int = "/".code;
+	/**
+		Provide an auto-complete on Enter/Tab key and command completion hints.
+	**/
 	public var autoComplete : Bool = true;
 
+	/**
+		Create new Console instance using provided font and parent.
+	**/
 	public function new(font:h2d.Font,?parent) {
 		super(parent);
 		height = Math.ceil(font.lineHeight) + 2;
@@ -74,12 +101,26 @@ class Console #if !macro extends h2d.Object #end {
 		addAlias("?", "help");
 	}
 
-	public function addCommand( name, ?help, args, callb : Dynamic ) {
+	/**
+		Add new command to console.
+		@param name Command name
+		@param help Optional command description text.
+		@param args An array of command arguments. 
+		@param callb The callback method taking the arguments listed in `args`.
+	**/
+	public function addCommand( name, ?help, args : Array<ConsoleArgDesc>, callb : Dynamic ) {
 		commands.set(name, { help : help == null ? "" : help, args:args, callb:callb } );
 	}
 
 	#end
-
+	
+	/**
+		Add new command to console via macro.
+		Only following callback parameters are supported: `Int`, `Float`, `String` and `Bool`.
+		@param ethis An expression that points at console instance.
+		@param name A String expression of the command name
+		@param callb An expression that points at the callback method.
+	**/
 	public macro function add( ethis, name, callb ) {
 		var args = [];
 		var et = haxe.macro.Context.typeExpr(callb);
@@ -106,10 +147,18 @@ class Console #if !macro extends h2d.Object #end {
 
 	#if !macro
 
+	/**
+		Add an alias to the existing command.
+		@param name Command alias
+		@param command Full command name to alias.
+	**/
 	public function addAlias( name, command ) {
 		aliases.set(name, command);
 	}
 
+	/**
+		Executes the String in same way the user would execute it.
+	**/
 	public function runCommand( commandLine : String ) {
 		handleCommand(commandLine);
 	}
@@ -179,10 +228,16 @@ class Console #if !macro extends h2d.Object #end {
 		}
 	}
 
+	/**
+		Checks if Console is currently shown.
+	**/
 	public function isActive() {
 		return bg.visible;
 	}
 
+	/**
+		Hides the Console.
+	**/
 	public function hide() {
 		bg.visible = false;
 		tf.text = "";
@@ -190,6 +245,9 @@ class Console #if !macro extends h2d.Object #end {
 		tf.cursorIndex = -1;
 	}
 
+	/**
+		Hides the shows and focuses the Console.
+	**/
 	public function show() {
 		bg.visible = true;
 		tf.focus();
@@ -409,6 +467,11 @@ class Console #if !macro extends h2d.Object #end {
 		}
 	}
 
+	/**
+		Print to console.
+		@param text The text to show in the log message.
+		@param color Optional custom text color.
+	**/
 	public function log( text : String, ?color ) {
 		if( color == null ) color = tf.textColor;
 		var oldH = logTxt.textHeight;
--- a/h2d/Console.hx
+++ b/h2d/Console.hx
@@ -6,12 +6,24 @@ import hxd.Key;
 	The console argument type.
 **/
 enum ConsoleArg {
+	/**
+		An integer parameter.
+	**/
 	AInt;
+	/**
+		A floating-point parameter.
+	**/
 	AFloat;
+	/**
+		A text string parameter.
+	**/
 	AString;
+	/**
+		A boolean parameter. Can be `true`, `false`, `1` or `0`.
+	**/
 	ABool;
 	/**
-		`AEnum` type acts as an `AString` but restricts allowed values to the provided list of strings.
+		A text string parameter with limitation to only accept the specified list values.
 	**/
 	AEnum( values : Array<String> );
 }
--- a/h2d/Console.hx
+++ b/h2d/Console.hx
@@ -38,7 +38,7 @@ typedef ConsoleArgDesc = {
 /**
 	A simple debug console integration.
 
-	Console can be focused manually trough `Console.show` and `Console.hide` methods
+	Console can be focused manually through `Console.show` and `Console.hide` methods
 	as well as by pressing the key defined by `Console.shortKeyChar`.
 
 	It's possible to log messages to console via `Console.log` method.
@@ -134,7 +134,7 @@ class Console #if !macro extends h2d.Object #end {
 	#end
 	
 	/**
-		A macro method to add a new command to console.
+		Add a new command to console. <span class="label">Macro method</span>
 
 		The `callb` method arguments are used to determine console argument type and names. Due to that, 
 		only the following callback argument types are supported: `Int`, `Float`, `String` and `Bool`.
--- a/h2d/Console.hx
+++ b/h2d/Console.hx
@@ -3,7 +3,7 @@ package h2d;
 import hxd.Key;
 
 /**
-	Console argument type.
+	The console argument type.
 **/
 enum ConsoleArg {
 	AInt;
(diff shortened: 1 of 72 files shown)


h2d/Console.hx
```diff
@@ -179,17 +228,26 @@ class Console #if !macro extends h2d.Object #end {
...
+	**/
 	public function hide() {
 		bg.visible = false;
 		tf.text = "";
 		hintTxt.text = "";
 		tf.cursorIndex = -1;
 	}
 
+	/**
+		Hides the shows and focuses the Console.
```
dpomier, line 249:
```suggestion
		Shows and focuses the Console.
```

h2d/Drawable.hx
```diff
@@ -13,12 +13,14 @@ class Drawable extends Object {
 
 	/**
 		By enabling smoothing, scaling the object up or down will use hardware bilinear filtering resulting in less crisp aspect.
```
dpomier, line 15:
```suggestion
		By enabling smoothing, scaling the object up or down will use hardware bilinear filtering resulting in a less crisp aspect.
```

h2d/Drawable.hx
```diff
@@ -13,12 +13,14 @@ class Drawable extends Object {
...
-		By default smooth is null and then Scene.defaultSmooth value is used.
+		By default smooth is `null` in which case `Scene.defaultSmooth` value is used.
 	**/
 	public var smooth : Null<Bool>;
 
 	/**
-		By enabling tile wrapping, you can have tiles which size exceed the texture size and will repeat instead of displaying clamped pixels.
+		By enabling tile wrapping, you can have tiles which size exceed the texture size and will repeat instead of displaying clamped pixels.  
+		tileWrap does not wrap within `Tile` bounds but withing underlying `Texture` and due to that, rendering a region of a texture with
+		tileWrap enabled won't cause that region to be tiled, but will extend into out-of-tile-bounds areas.
```
dpomier, line 23:
```suggestion
		`tileWrap` does not wrap within `Tile` bounds but withing underlying `Texture` and due to that, rendering a region of a texture with
		`tileWrap` enabled won't cause that region to be tiled, but will extend into out-of-tile-bounds areas.
```

h2d/Drawable.hx
```diff
@@ -91,7 +93,8 @@ class Drawable extends Object {
 	}
 
 	/**
-		Set the `colorMatrix` value by specifying which effects to apply. Calling adjustColor() reset the colorMatrix to `null`
+		Set the `colorMatrix` value by specifying which effects to apply.
+		Calling `adjustColor()` without arguments will reset the colorMatrix to `null`
```
dpomier, line 97:
```suggestion
		Calling `adjustColor()` without arguments will reset the colorMatrix to `null`.
```

h2d/Drawable.hx
```diff
@@ -164,7 +167,8 @@ class Drawable extends Object {
 	}
 
 	/**
-		Add a shader to the object shaders
+		Add a shader to the object shaders
```
dpomier, line 170:
```suggestion
		Add a shader to the object shaders.
```

h2d/Dropdown.hx
```diff
@@ -30,22 +30,54 @@ private class Fake extends Object {
 	}
 }
 
+/**
+	`h2d.Dropdown` is a simple UI component that creates an interactive drop-down list.
+	Dropdown will add an `h2d.Flow` to `Scene` when opening in order to be visible above other objects. See `dropdownLayer`.  
+	Note that when dropdownList opens and closes, item objects will recieve the `onHierarchyChanged` callback.
```
dpomier, line 36:
```suggestion
	Note that when `dropdownList` opens and closes, item objects will receive the `onHierarchyChanged` callback.
```

h2d/Dropdown.hx
```diff
@@ -30,22 +30,54 @@ private class Fake extends Object {
...
+**/
 class Dropdown extends Flow {
 	var items : Array<h2d.Object>;
 	var fake : Fake;
 	var cursor : h2d.Bitmap;
 	var arrow : h2d.Bitmap;
 
+	/**
+		A background Tile that is shown when user hover over an item in the dropdown list.
```
dpomier, line 45:
```suggestion
		A background Tile that is shown when the user overs an item in the dropdown list.
```

h2d/Dropdown.hx
```diff
@@ -30,22 +30,54 @@ private class Fake extends Object {
...
+	**/
 	public var tileOverItem(default, set) : h2d.Tile;
+	/**
+		A Tile used to visualize an arrow of the dropdown when list is closed.
```
dpomier, line 49:
```suggestion
		A Tile used to visualize an arrow of the dropdown when the list is closed.
```

h2d/Dropdown.hx
```diff
@@ -30,22 +30,54 @@ private class Fake extends Object {
...
+	**/
 	public var tileArrow(default, set) : h2d.Tile;
+	/**
+		A Tile used to visualize and arrow of the dropdown when list is open.
```
dpomier, line 53:
```suggestion
		A Tile used to visualize and arrow of the dropdown when the list is open.
```

h2d/Dropdown.hx
```diff
@@ -30,22 +30,54 @@ private class Fake extends Object {
...
+	**/
 	public var tileArrowOpen : h2d.Tile;
-
+	
+	/**
+		When disabled, user would not be able to change selected item. ( default : true )
```
dpomier, line 58:
```suggestion
		When disabled, the user would not be able to change the selected item. ( default : true )
```

h2d/Dropdown.hx
```diff
@@ -30,22 +30,54 @@ private class Fake extends Object {
...
+	**/
 	public var canEdit(default,set) : Bool = true;
+	/**
+		A reference to `h2d.Flow` that will contain items.
+		Adding objects to this Flow will not automatically add them to item list, see `addItem`.
```
dpomier, line 63:
```suggestion
		Adding objects to this Flow will not automatically add them to the item list, see `addItem`.
```

h2d/Dropdown.hx
```diff
@@ -30,22 +30,54 @@ private class Fake extends Object {
...
+	**/
 	public var dropdownList : Flow;
+	/**
+		A Scene layer to which dropdownList will be added when opening dropdown. ( default : 0 )
```
dpomier, line 67:
```suggestion
		A Scene layer to which `dropdownList` will be added when opening dropdown. ( default : 0 )
```

h2d/Dropdown.hx
```diff
@@ -30,22 +30,54 @@ private class Fake extends Object {
...
+		Currently selected item index. To deselect an item, set it to `-1`.
+	**/
 	public var selectedItem(default, set) : Int = -1;
+	/**
+		Currently highlighted item index.
+	**/
 	public var highlightedItem(default, null) : Int = -1;
-
+	/**
+		When enabled, dropdown list will appear above the dropdown.
```
dpomier, line 79:
```suggestion
		When enabled, the dropdown list will appear above the dropdown.
```

h2d/Dropdown.hx
```diff
@@ -157,6 +189,9 @@ class Dropdown extends Flow {
 		return tileOverItem = t;
 	}
 
+	/**
+		Adds an Object to the dropdown list. Object is not restricted to be same type across all items.
```
dpomier, line 193:
```suggestion
		Adds the Object `s` to the dropdown list. `s` is not restricted to be the same type across all items.
```

h2d/Font.hx
```diff
@@ -77,14 +107,33 @@ enum FontType {
...
 
 class Font {
-
 	public var name(default, null) : String;
+	/**
+		Current font size. Font can be resized with `resizeTo`.
+	**/
 	public var size(default, null) : Int;
+	/**
+		The baseline value of the font represent the base on which characters will sit.
```
dpomier, line 116:
```suggestion
		The baseline value of the font represents the base on which characters will sit.
```

h2d/Font.hx
```diff
@@ -77,14 +107,33 @@ enum FontType {
...
+		Font line height provides vertical offset for each new line of the text.
+	**/
 	public var lineHeight(default, null) : Float;
+	/**
+		Reference to the source Tile containing all glyphs of the Font.
+	**/
 	public var tile(default,null) : h2d.Tile;
 	public var tilePath(default,null) : String;
+	/**
+		The font type. BitmapFonts rendered as-is, but SDF fonts will use extra shader to produce scalable smooth fonts.
```
dpomier, line 130:
```suggestion
		The font type. BitmapFonts rendered as-is, but SDF fonts will use an extra shader to produce scalable smooth fonts.
```

h2d/Font.hx
```diff
@@ -77,14 +107,33 @@ enum FontType {
...
+	**/
 	public var type : FontType;
+	/**
+		Font charset allows to resolve specific char codes that are not directly present in glyph map as well as detect spaces.
+		Defaults to `hxd.Charset.getDefault()`
```
dpomier, line 135:
```suggestion
		Defaults to `hxd.Charset.getDefault()`.
```

Yanrishatum:
@dpomier Accepted some of suggestions (and I think some would be rephrased, as they are a bit unclear), but I'm not sure about overabundance of `the` additions. English is not my primary language, but with all those extra `the` it reads a lot more awkward for me, so I'm a bit hesitant on that part.

Gama11:
Not a native speaker either,  but all the `the` additions seem correct to me. It's not really a matter of preference, without it, it sounds wrong.

h2d/Dropdown.hx
HeapUnderfl0w: Shouldn't this be `when the user hovers over an item` instead of `when the user overs an item` ?

kennygoff:
Yeah, I'm a native english speaker and the changes are correct with `the`. I'm no linguist but it definitely sounds right and checking some articles seems to confirm that: https://owl.purdue.edu/owl/general_writing/grammar/using_articles.html

Yanrishatum:
As to not spam with Issues when I find questionable places while writing docs, I'm just going to note them here and ping @ncannasse for comments. :)

https://github.com/HeapsIO/heaps/blob/master/h2d/col/IPoint.hx#L32-L38
Is it supposed to return `Point` and not `IPoint` or copy-paste bug?

ncannasse:
Yes it's definitely a copy-paste bug :D

Yanrishatum:
I'll be temporarily hosting more up-to-date API docs on my server in case anyone needs it. https://wb.yanrishatum.ru/heaps/api/

ncannasse:
Here's some comments from a first look:

General

- there's no need IMHO for each class/type to quote it's name at the beginning of the documentation. For instance https://wb.yanrishatum.ru/heaps/api/h2d/Graphics.html , I think it will be better with "~h2d.Graphics provide~ Simple interface to draw arbitrary **2D** geometry." ; Maybe that's not what we used to do, but I think it's better while we're at it to makes things consistent. Also avoid "Is" or "Is used to" or "Provide", or "Implements" as a starting word. It should directly tell what it does, just like we do for methods and variables.

- for variable with default values, the default is already generated in documentation (see https://wb.yanrishatum.ru/heaps/api/h2d/Console.html, both static and shortKeyChar for example). No need then to have the default specified in text documentation. This will also prevent that when changing the default value we break the documentation.

Specific

- I'm not at all a fan of ascii art for expressing complex visual things such as alignments here https://wb.yanrishatum.ru/heaps/api/h2d/Align.html . These should be done either by linking to a manual page that contains images. For instance in that case, the Align class desc should be "Text alignment rules. For more details on how it's applied, see (this manual page)". We could add also for each field "see (this manual page) for more information".

- Console.add is a macro member function. The documentation generator does not mention this and the documentation should be updated.

- filter Ambient apply a color matrix using a mask to tell which parts are to be changed or kept (0...255) - invert tells if the mask should be the parts with ambient applied or the parts that are kept untouched (does 1.0-mask on the mask values) and power allows to control transition speed - pow(mask,power)

Otherwise I don't think I'll be reviewing all of the documentation before merging, as it's often better to have some documentation than none :)

Yanrishatum:
I'm marking this as ready, but I would suggest leaving it open for a bit (a week or so?), in case some crazy enough people could go over it and provide suggestions on grammar and still unclear areas. I won't do as massive doc PRs like that anymore and next API documentation updates are going to be smaller scale and target specific areas.
Also, here's current documentation coverage: https://wb.yanrishatum.ru/heaps/api/coverage.html
Other packages are yikes.

Yanrishatum:
Well, looks like nobody gonna review it, so it's up for merge.

ncannasse:
Sorrrrrry it's conflicting :'( can you fix so I can merge ?

Yanrishatum:
@ncannasse Done. Also, pls fix `h3d.shader.AlphaMap` (post https://github.com/HeapsIO/heaps/commit/865bf5b0b79f6c59c33230b0d70854a4dc93e89c ), it doesn't have `input.uv` declared and it broke documentation generator again ;)

ncannasse:
Done ! Thanks !

Yanrishatum:
Just don't forget to update https://heaps.io/api/ ;)

ncannasse:
Just triggered a rebuild of heaps-doc on Travis, let's see...

ncannasse:
Ok I actually need to make a new heaps release, and update HEAPS_VER
version on heaps.io repo

PR #844, opened 2020-06-29
Introduce BatchDrawState for h2d

### LINKED ISSUE #838: [RFC] Multi-Texture support for h2d batched renderers
The one-texture-per-batcher limitation always was annoying to deal with, since while in release version it provides benefits of batching all the elements in one draw-call taking from single atlas, during development it's sub-optimal, since depending on workflow those classes can become unusable due to artists using lots of different images instead of one atlas, and for quick iterations - rebuilding atlas after any little change is the last thing you'd want to do.
Hence I would propose introducing support for multiple textures in following classes: `Graphics`, `SpriteBatch` and `TileGroup`.

## Implementation
I already did an experimental implementation in Heeps: https://github.com/Yanrishatum/heeps/blob/master/ch2/BatchDrawer.hx#L397-L471
In essence, everything still resides in a singular buffer, but on top of Content, there will be a BatchState instance, that would hold information unique to specific batch sequence (Starting offset, tile and triangle count), which are then rendered with multiple draw-calls. Those batch instances are pooled as to minimize the allocation count.
Compared to experimental implementation I don't plan to introduce multiple textures in one draw-call as I did in Heeps, so one BatchState going to be restricted to one Texture.
As a side note: If texture switched, it makes new state regardless if that texture was used in previous states. This is for obvious reason of render ordering. Also experimental implementation is restricted to quads, which is not compatible with regular TileGroup and Graphics, hence proper implementation would use different counters.

## Drawbacks
* Optimization aspect is gone to a degree and heavily depends on how those classes are utilized.
  * Should not affect performance if used the old-fashioned way.
* Slight overhead and memory footprint between Content and Object.
* Causes multiple `beginDrawObject` calls.
  * Can be mitigated by introducing `RenderContext.drawBatchStates(object, states)` or another way to skip some of `beginDrawObject` setup and just switch textures around.

## Positives
* Ability to use multiple textures in batches in exchange for more drawcalls
* Faster iteration time, since having singular atlas is no longer mandatory.
* It's possible to later compile all images to singular atlas and get benefits of less drawcalls on release, but faster iteration time during development.
* Also in case when content doesn't fit one atlas, it will be possible to use multiple atlases.

## Extra notes
* Other approach is to use atlas that builds at runtime, and again, I have an experimental implementation: https://github.com/Yanrishatum/heeps/blob/master/cherry/RuntimeAtlas.hx
* Multi-texture support will allow to start implementation of #453, bringing full support of bmf format.
* Also will allow to simplify HtmlText a lot, since it'll be possible to cram everything into single Content instance instead of awkward objects added to instance. (And draw count will remain the same, but less unnecessary overhead of multiple objects)

## Things to discuss
* Should unused BatchState instances be stored in global pool or individual pools?
* Should used states be held in Content or Object? I'm leaning towards Content, as to allow direct usage of Content methods and not bother with forgetting to increase counters or change to new state.
* Should RenderContext be aware of batchstates and allow simplified state switch? I'd say yes.

@ncannasse :)

Yanrishatum:
Resolve #838 
* Introduced `h2d.RenderTexture.BatchDrawState` class.
* Integrated said class to `TileGroup`, `Graphics` and `SpriteBatch`, enabling support for multiple textures in said classes.
* Added simple DrawingTiles sample.

Public API and behavior should remain unchanged and when used with optimized approach of one texture overhead should not be noticeable.


ncannasse:
Thanks !

Once thing that bother me is that each class that wants to deal with BatchState have to deal with manual chaining etc. 

I think a more high level API where we just create a single BatchStates class instance and emit the tile on it that will maintain its inner list would be easier to deal with. I would implement its own pooling so you that SpriteBatch implementation is much more easy (no need for its own pseudo-pooling).

Also, I think for these we might want to have a h2d.impl package for this kind of "tools" classes that are not directly part of the API, but used by it.

Yanrishatum:
Unifying usage of BatchState on each class would be difficult, since SpriteBatch and TileGroup use quads and graphics use tris, leading to different counters. On top of that, TileGroup implements range limiter, SpriteBatch uses custom buffer instead of separate primitive and Graphics implements multi-buffer rendering.
Having `emitTile` won't help, because that's incompatible with Graphics. I do agree it's rather cumbersome to use, but to mitigate it, h2d would have to let DrawState handle much more than just "I hold that texture", namely allow it to work as a common Primitive that handles the buffers as well as make it work in both quad and tri modes.

ncannasse:
I didn't mean to have this as a primitive. The buffers would still be
managed by the classes directly. So not that many changes compared to what
you did, just a BatchStates that manage its own private next instead of
being accessed/linked manually by classes using it.

Le sam. 4 juil. 2020 à 13:20, Pavel Alexandrov <notifications@github.com> a
écrit :

Yanrishatum:
Moved the class to `h2d.impl` package and reimplemented it to self-contained class. I opted to move rendering to the batch class to ensure uniform rendering. 
Also added Graphics+tiles usage sample to DrawingTiles.

ncannasse:
Few comments:
- we should review the semantics of  next/nextTile/add : let's have setTile(t)/setTexture(t) then add(count) as two separate functions, which makes it more easy
- I also don't see much the point of having a texture passed as BatchDrawState constructor or not clearing the textures on clear(), I suggest we always start with null texture + set to null on clear, unless there's something I've missed.
- I'm not sure it's worth calculating and caching the offsets into each BatchDrawState element. It makes things more complicated and only saves a few operations when drawing. So what about recalculating things when drawing directly ?

Yanrishatum:
> we should review the semantics of next/nextTile/add : let's have setTile(t)/setTexture(t) then add(count) as two separate functions, which makes it more easy

> I also don't see much the point of having a texture passed as BatchDrawState constructor or not clearing the textures on clear(), I suggest we always start with null texture + set to null on clear, unless there's something I've missed.

Done.
It was passed on constructor due to Graphics splitting its buffers internally, and I needed to preserve the previous buffer active texture. If it's set after instancing, it just becomes a bit more cumbersome:
```haxe
var tex = state.currentTexture;
state = new BatchDrawState(false);
state.setTexture(tex);
```

> I'm not sure it's worth calculating and caching the offsets into each BatchDrawState element. It makes things more complicated and only saves a few operations when drawing. So what about recalculating things when drawing directly ?

Redesigned the state to a degree. Along with removal of precalc of elements, also removed tracking of offset, since it always starts from 0 anyway, and there's no need to hold `isQuads`, since it was used for precalc, but since quads and tris use their own methods - they already calculate it their own way. Tracking of total added vertices is important to properly calculate the length value if `offset` is not 0, but `length` is -1.
Also added more expansive documentation while I was at it.

ncannasse:
Thanks !

PR #856, opened 2020-07-14
Fix name collision in Pak FS

Yanrishatum:
Fixes name collision introduced by https://github.com/HeapsIO/heaps/commit/37414f0189b317b08d982b2dfc00072a4eafa190 on non-system targets.
This way, it should no longer break the travis status as well as fixes usage of pak files on HTML5.


ncannasse:
I think I've somewhat fixed it in 904199b00c14865a9f7479b82aaad2427e38fb93

PR #863, opened 2020-07-31
Render To Texture with depth and stencil.

lusius:
When creating a combined depth-stencil buffer using the heaps Depth24Stencil8, WebGL GL_DEPTH_STENCIL is passed for creating the buffer. 

Setting this buffer both in depth and stencil attachment will render errors on mobile devices running webgl1. 

This fix binds the combined buffer to the DEPTH_STENCIL_ATTACHMENT.

Since the only way to have stencil is using the Depth24Stencil8 format the setting of stencil the old way should never be reached but keeping for safety reasons and maybe future two separate buffers will actually be created.


lusius:
ping @ncannasse

PR #864, opened 2020-08-04
Fix #862

### LINKED ISSUE #862: [h2d.Layers] Inconsistent addChildAt and getChildIndex
```haxe
class Main extends hxd.App {
	override function init() {
		new h2d.Object(s2d);
		new h2d.Object(s2d);
		
		var obj = new h2d.Object();
		s2d.addChildAt(obj, 1);
		trace(s2d.getChildIndex(obj));
	}
	static function main() {
		new Main();
	}
}
```

This will print 2 but expected 1, because we called addChildAt 1

Yanrishatum:
What the title says. Resolves Layers inconsistent behavior in #862
* `addChild` and `addChildAt` now operate same way as Object with exception of adding to the default layer.
* `add` now can take additional child index.
* `getChildIndex` now returns child index local to the layer, not `children` list. 
* `getChildAt` now returns child at default layer. 
* Added `getChildAtLayer` which allows to specify the lookup layer.
* Added `defaultLayer`, which dictates the `addChild`, `addChildAt` and `getChildAt` default layer.
I can remove defaul layer feature and just default everything to 0 if needed.

Breaking changes:
`addChildAt` no longer adds to a layer, but rather adds to default layer at given index. This is a major breaking change for obvious reasons. 
`getChildIndex` returns index local to the layer, which may lead to bugs in places that rely on it. 
`getChildAt` now looks up only at default layer and `getChildAtLayer` should be used instead.

Alternatively, it's possible to revert last two changes and add `getChildLayerIndex` with `getChildAtLayer` with `getChildIndex` / `getChildAt` to behave as usual or throw an error asking to utilize Layers-specific methods.


Yanrishatum:
More detailed justification on `defaultLayer` as a sample:
```haxe
for (bg in backgrounds) new Bitmap(bg, layers);
layers.defaultLayer = 1;
for (obj in objects) new GameObject(obj, layers);
new Player(layers);
layers.defaultLayer = 3;
for (fg in foregrounds) new Bitmap(fg, layers);
new Effects(layers);
layers.defaultLayer = 4;
new GameUI(layers);
new Text(font, layers).text = Game.version;
new DebugUI(layers);
```
instead of awkward instantiate -> store to variable -> do `layers.add` or `layers.add(new Instance(), layer);`. Basically allows to utilize the passing of parent in the constructor. As a downside: If forgotten, can lead to unexpected "bugs", but at this point I would blame the programmer for not paying attention.

Also I think it would be better for `addChildAt` to check if child is already added to the Layers and instead of re-adding it to default layer - move it withing its current layer.

ncannasse:
We need testing in both NG and DB. Ping @trethaller

Yanrishatum:
Resolved merge conflict + fixed a few bugs + added `domkit` check.

trethaller:
Ran very basic tests on NG & DB, changes don't break anything obvious. In DB layers are not used except for a few s2d.add/addChildAt

ncannasse:
I don't agree with the defaultLayer thing, and sorry as I just noticed you commented that proposal back in July and I didn't reply.

I think getChildAt / getChildIndex should still work on all children, not regarding on which layer they are, the same addChildAt is. This way we don't break anything wrt Object specification.

So the fix is only to keep addChildAt the same as h2d.Object

Yanrishatum:
May I know the reasoning why you don't accept `defaultLayer`? 

> I think getChildAt / getChildIndex should still work on all children, not regarding on which layer they are, the same addChildAt is. This way we don't break anything wrt Object specification.

I would still prefer to keep layer-specific methods, even if renamed.

ncannasse:
I want addChild to add to topmost as it currently does, so it keeps Object
behavior.
So we don't need defaultLayer, as we only have  add()  and I prefer if the
layer is passed explicitly here instead of hidden into some state variable
that might be modified by several sources and lead to bugs because it's not
at the value you expected it to be.

h2d/Layers.hx
```diff
@@ -23,45 +23,66 @@ class Layers extends Object {
...
+		// Populate layer list
+		while ( layer >= layerCount )
 			layersIndexes[layerCount++] = children.length;
-		super.addChildAt(s,layersIndexes[layer]);
-		for( i in layer...layerCount )
+		
+		if ( index != -1 ) {
+			// Prevent inserting out of layer bounds.
+			if ( layer == 0 )
+				super.addChildAt(s, hxd.Math.imin(layersIndexes[0], hxd.Math.imin(index, layersIndexes[layer])));
```
ncannasse, line 58: There's two imin on layersIndexes[0] here, can you fix ?

ncannasse:
Last review and one fix left before merge ;)

PR #868, opened 2020-08-24
add option to enable buffering for emitTile

ZwodahS:
Add option "heaps_buffering" to enable buffering in emitTile.

### DIFF
--- a/h2d/RenderContext.hx
+++ b/h2d/RenderContext.hx
@@ -12,7 +12,7 @@ private typedef RenderZoneStack = { hasRZ:Bool, x:Float, y:Float, w:Float, h:Flo
 @:access(h2d.Scene)
 class RenderContext extends h3d.impl.RenderContext {
 
-	static inline var BUFFERING = false;
+	static inline var BUFFERING = #if heaps_emit_tile_buffering true #else false #end;
 
 	public var globalAlpha = 1.;
 	public var buffer : hxd.FloatBuffer;


ncannasse:
This is a quite old option that hasn't been tested in years. Does it still works ? :D

Yanrishatum:
It does, and from what I know from implementation - should work fairly reliably, since `beginDraw` does thorough checks and flushes if there's any deviations like shader composition change, texture swap or filter shenanigans. And I don't see any obvious fault positions from what I know (there may be edge cases, but should work fine for most cases).

However I don't like the naming of the switch, as it's not very intuitive what kind of buffering it enables. While it's longer, but I'd rename it to `heaps_emit_tile_buffering` or something like that. 
In general, I see no problem giving ability to enable it with a compilation flag. Since it's gated behind compilation flag - it's basically "use at your own risk", same with enabling MP3 in Res completion and such.

ncannasse:
Agreed that heaps_buffering is too "generic"

ZwodahS:
I've been testing this for the last 2-3 weeks, doesn't seem to have problem. Was just curious why there wasn't a flag for this. 

Will change the flag and rebase onto the new master.

ZwodahS:
Additional context as to why I investigated this.

I needed some form of layer in SpriteBatch like h2d.Layers, and was going to implement it but found that I could just use Bitmaps and Layers with this flag turn on.

PR #870, opened 2020-08-26
Added documentation for the hxd.Rand class

haddock7:


ncannasse:
The documentation should be improved with the following:
- hxd.Rand is a seeded random number generator, that allows to get always the same results starting from a given seed.
- srand() returns a random float between -scale and +scale (excluded)

haddock7:
@ncannasse I have modified the documentation with your suggestions.

ncannasse:
Thanks

PR #871, opened 2020-08-27
Add support for Home button on game controllers

jstolarek:
It's been annoying me that there's no support for Home button on game controllers in heaps. It's a trivial fix.

### DIFF
--- a/hxd/Pad.hx
+++ b/hxd/Pad.hx
@@ -41,6 +41,7 @@ typedef PadConfig = {
 	LT : Int,
 	RT : Int,
 	back : Int,
+	home : Int,
 	start : Int,
 	analogClick : Int,
 	ranalogClick : Int,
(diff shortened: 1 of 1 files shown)


Yanrishatum:
You also should add home button definition to other pad configs, otherwise you introduce compilation errors on everything except HLSDL.

ncannasse:
Given Home is specific to Xbox controller and is not always customizable (consoles etc.) I don't think it should be part of default configuration.

jstolarek:
> Home is specific to Xbox controller

I don't believe Home is specific to XBox - it is also present on Playstation controllers and PC controllers. I have that button on Logitech F310 and F710, Hori Fighting Commander (primarily a PS4 controller) and Mayflash F101 arcade stick (compatible with essentially everything - PC, PS4, Switch, XBox).

PR #874, opened 2020-08-31
Rename startDrag to startCapture

### LINKED ISSUE #872: Misleading name of `startDrag`
As title says, `startDrag` is a bit misleading, since proper name would be `captureEvents`. When seeing current name, people just assume that they would start the drag operation and receive the mouse movement events, which is not true, as all input events are now delegated to the callback, hence the more proper name - `captureEvents`.
Essentially what I propose:
1. Rename the `startDrag` method to `captureEvents`, and `stopDrag` to `stopCapture` respectively.
2. Leave old start/stop methods with a deprecation notice.
This way, it won't cause a breaking change, but will mitigate the misleading naming of the method. And will add an ample amount of time for people to fix it, because deprecation warnings are annoying ;)

As a side note: In the h2d API docs I will call it as "capture input events", not as "drag and drop operation".

Yanrishatum:
Resolve #872 
* Renamed `startCapture` and `stopCapture` in `hxd.SceneEvents`, `h2d.Scene` and `h2d.Interactive`.
* Old names are not marked as deprecated.
* Also added `refEvent` argument to `Interactive.startCapture` to introduce feature parity with the SceneEvents/Scene.
* I didn't touch the documentation, since I will touch it up anyway when merging with API docs PR.

I have a question, however. Is there a point in Scene and Interactive capture methods not being inline? Well, with `Interactive.startCapture` it more or less makes sense, but not so much for others, since they just delegate it to `SceneEvents` and cause extra method calls for no reason.


ncannasse:
For touch events, I prefer that we directly pass touchId to startCapture.

Yanrishatum:
Changed reference event to direct `touchId:Int`, but left `stratDrag` untouched as to not introduce breaking changes.

ncannasse:
Uhm.... I don't see where events get filtered by touchId...

Yanrishatum:
https://github.com/HeapsIO/heaps/blob/a19298de72f2b80ac9eede11af57f89e5b301e09/hxd/SceneEvents.hx#L355-L363
@ncannasse Here? :)

Yanrishatum:
Resolved merge conflict from h2d docs addition.

ncannasse:
Thanks !

PR #876, opened 2020-09-10
Fix #875

### LINKED ISSUE #875: BatchDrawState in Graphics
There's an issue wrt #844 : you can draw something in h2d.Graphics without to specify a texture, and currently it BatchDrawState will use the null texture which will be 0xFF00FF instead of white.

Ping @Yanrishatum

Yanrishatum:
* Fixes #875 as mentioned by introducing separate white tile when doing line or solid color fills. 
* Fixed an oversight with previous drawbatchstate implementation where passing null tile to `beginTileFill` was not allowed, yet in original version it would use current `Graphics.tile` value.
* Also fixed an oversight in state switch logic: Not changing between textures but not adding any geometry should not produce new state instances.

### DIFF
--- a/h2d/Graphics.hx
+++ b/h2d/Graphics.hx
@@ -1,4 +1,5 @@
 package h2d;
+import h2d.impl.BatchDrawState;
 import hxd.Math;
 
 private typedef GraphicsPoint = hxd.poly2tri.Point;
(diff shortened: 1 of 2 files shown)


ncannasse:
Need resolve ;)

Yanrishatum:
Note to self: Don't use github merge resolver, it seem to screw up file line endings.

Yanrishatum:
Resolved merge conflict (dunno why it couldn't resolve it).

ncannasse:
Why introduce `fillTile` as a member var while it's a constant ? I think it's better reassigning tile to 0xFFFFFFFF each begin.

Yanrishatum:
@ncannasse To avoid reallocation of Tile instance for single use. Removed.

PR #881, opened 2020-09-23
Breaking line japanese rules

Azrou:
Mark Chiisai Japanese characters as complementChar that must not be broken for a new line

### DIFF
--- a/h2d/Text.hx
+++ b/h2d/Text.hx
@@ -217,7 +217,8 @@ class Text extends Drawable {
 			var e = font.getChar(cc);
 			var newline = cc == '\n'.code;
 			var esize = e.width + e.getKerningOffset(prevChar);
-			if( font.charset.isBreakChar(cc) ) {
+			var nc = text.charCodeAt(i+1);
+			if( font.charset.isBreakChar(cc) && (nc == null || !font.charset.isComplementChar(nc)) ) {
 				if( lines.length == 0 && leftMargin > 0 && x > maxWidth ) {
 					lines.push("");
 					if ( sizes != null ) sizes.push(leftMargin);
@@ -236,7 +237,8 @@ class Text extends Drawable {
 					var e = font.getChar(cc);
 					size += e.width + letterSpacing + e.getKerningOffset(prevChar);
 					prevChar = cc;
-					if( font.charset.isBreakChar(cc) ) break;
+					var nc = text.charCodeAt(k+1);
+					if( font.charset.isBreakChar(cc) && (nc == null || !font.charset.isComplementChar(nc)) ) break;
 				}
 				if( size > maxWidth || (!breakFound && size + afterData > maxWidth) ) {
 					newline = true;
--- a/hxd/Charset.hx
+++ b/hxd/Charset.hx
@@ -138,6 +138,15 @@ class Charset {
 		return isSpace(code) || isCJK(code);
 	}
 
+	static var complementChars : Map<Int,Bool> = {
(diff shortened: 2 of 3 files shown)


ncannasse:
This PR includes also vertical align commit, please update ;)

h2d/Text.hx
```diff
@@ -236,11 +237,12 @@ class Text extends Drawable {
...
 					size += e.width + letterSpacing + e.getKerningOffset(prevChar);
 					prevChar = cc;
-					if( font.charset.isBreakChar(cc) ) break;
+					var nc = text.charCodeAt(k+1);
+					if( font.charset.isBreakChar(cc) && (nc == null || !font.charset.isComplementChar(nc)) ) break;
 				}
 				if( size > maxWidth || (!breakFound && size + afterData > maxWidth) ) {
 					newline = true;
-					if( font.charset.isSpace(cc) ){
+					if( font.charset.isBreakChar(cc) && (nc == null || !font.charset.isComplementChar(nc)) ){
```
Yanrishatum, line 245: It should not check for `isBreakChar`, and will introduce issues. It checks for space in order to trim whitespace after performing line break, since regular space have to be removed, but not renderable character (i.e. Chinese/Japanese characters). With your change, it will remove the characters it shouldn't, when there's no complement character present.

Yanrishatum (changes requested):
Additionally, same changes have to be made to HtmlText, since it uses custom line-break implementation.

ncannasse:
@Azrou can you check @Yanrishatum comment and update the PR ?

h2d/Text.hx
Azrou: Indeed. I pulled that from an older and modified heaps branch, it's irrelevant here.
I added complement chars to HtmlText too.

Thanks for review !

Azrou:
@ncannasse Done :-)

hxd/Charset.hx
```diff
@@ -138,6 +138,15 @@ class Charset {
...
 	}
 
+	static var complementChars : Map<Int,Bool> = {
+		var str = "ヽヾーァィゥェォッャュョヮヵヶぁぃぅぇぉっゃゅょゎゕゖㇰㇱㇲㇳㇴㇵㇶㇷㇸㇹㇺㇻㇼㇽㇾㇿ々〻";
+		[for( i in 0...str.length ) str.charCodeAt(i) => true];
+	}
+
+	public function isComplementChar(code) {
+		return complementChars.exists(code);
+	}
```
Yanrishatum, line 148:
```suggestion
	static var complementChars : Array<Int> = {
		var str = "ヽヾーァィゥェォッャュョヮヵヶぁぃぅぇぉっゃゅょゎゕゖㇰㇱㇲㇳㇴㇵㇶㇷㇸㇹㇺㇻㇼㇽㇾㇿ々〻";
		[for ( i in 0...str.length ) str.charCodeAt(i)];
	}

	public function isComplementChar(code) {
		return complementChars.indexOf(code) != -1;
	}

```
There's no point in having a Map here, only increases memory footprint for no reason, since it contains only chars that would prevent line breaks and nothing else, especially since check only tests for value being here, not whether its `true` or `false`.

ncannasse: I was the one suggesting Map , so we get a lot faster check, given it can gets called a lot.

Yanrishatum: I suppose it's due to IntMap using different lookup approach? (something like binary tree?) I'm curious why Map is faster in that case.

ncannasse:
Yes Map uses tree-like lookup, it's implementation dependent but typically
lookup is O(log n) instead of O(n)

Yanrishatum:
Got it. Good to know. Thanks.

PR #890, opened 2020-10-06
Fix Float32Expand returning invalid length

nspitko:
Dug this up while bug chasing during LD. The JS impl returns `pos` instead of `array.length`, which effectively breaks it outside of specific access patterns.


nspitko:
This actually doesn't fully solve the issue; the purpose of `pos` is to track length internally so you can pre-alloc buffers larger than are actually used.

The pattern still fails when trying to array access the buffer past it's pre-allocated bounds instead of using push. This is just generally broken on JS. 

Closing for now, will either work out a better fix if I have time, or file an issue.

PR #891, opened 2020-10-06
Implement @doc meta for shaders

Yanrishatum:
Adds support for `@doc` metadata in HXSL shaders. 
Primary aim is to enable documentation of the variables generated by macros.
Usage example (from Base2d):
```haxe
		@doc("Discard pixels with alpha below 0.001")
		@const var killAlpha : Bool;
```

### DIFF
--- a/hxsl/Ast.hx
+++ b/hxsl/Ast.hx
@@ -83,6 +83,7 @@ enum VarQualifier {
 	Range( min : Float, max : Float );
 	Ignore; // the variable is ignored in reflection (inspector)
 	PerInstance( v : Int );
+	Doc( s : String );
 }
 
 enum Prec {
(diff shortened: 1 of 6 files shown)


ncannasse:
I prefer if we have doc as part of the qualifiers Doc( s : String ) as this reduces the implementation overhead and possibility of loss of data when dealing with it.

Yanrishatum:
Sure, I'll rework it. I actually at first considered putting it into qualifiers, but that would put extra data into cache and such, which isn't particularly necessary due to it being useful only for completion and api docs purposes.

Yanrishatum:
Done. Also gated `hxsl.Macros` code in `#if macro` to make compilation server happy in regard to accessing macro context stuff.

ncannasse:
Not sure why but some commas were removed on unrelated lines ;) can you fix so I can merge ?

Yanrishatum:
Restored. They were removed when I was deleting the `doc:` declarations, and removed redundant commas out of habit.

PR #902, opened 2020-11-02
Tiny docstring fix for Point.scale()

zommerfelds:
Nit about not actually "returning" anything :)


Yanrishatum:
Looks like an oversight after math types rework https://github.com/HeapsIO/heaps/commit/056f294528fb956e980ba006a082d807fc80f8b8 which made `scale` localized.

PR #903, opened 2020-11-03
Implement 2d filter resolution scaling.

### LINKED ISSUE #885: [RFC] Resolution scaling in filters
Currently, filters use static resolution based on the filtered object boundaries. While it makes sense on paper, on practice it have a specific unaccounted flaw in the rendering pipeline. I would like to propose a fix to that.

# Problem
* Using `Scene.scaleMode` causes filters to use internal Scene resolution, and for pixel-art games that want to use non-pixel-perfect effects this is detrimental, and additionally introduces less smooth subpixel movement, since it's rendered onto a small resolution texture. For example, 320x240 internal resolution would pretty much render filters useless, since it would use very small textures, and effects like glow or blur would be of very bad quality. 
* Additionally, it's impossible to control in general the resolution of the textures. For example, it's not always preferable to use full-scale texture for something, or sometimes it would be better to upscale the resolution to achieve better fidelity.

# Proposal
Introduce two new variables to `h2d.filters.Filter`:
* `public var resolutionScale:Float = 1;` which would control the resulting texture resolution and act as a multiplier. That would give use the control over texture resolution, while retaining automatic size calculation for said texture.
* `public var useScreenResolution:Bool = false;` which would disable the resolution scaling, but in turn will use the window resolution as a base for the texture resolution, instead of internal Scene resolution. This would allow to have same visual fidelity of the filtered object on small internal resolution of the Scene while avoiding manual calculation of the resolution scaling. Plus it would use non-uniform width and height, since scaleMode can stretch either one.

# Implementation notes
Mainly it would affect resolution calculation and what gets assigned to `filterMat` in the shader and it should not affect the current behavior of the filters unless explicitly used.

Yanrishatum:
Resolves #885

# Changes overview

* `Object.clipBounds` now expects `scaleX` and `scaleY` arguments in order to properly calculate boundaries mask with respect to filter resolution scaling.
* Moved `if (maxExtent >= 0)` check inside `autoBounds`, because there's no point in having it ouside, since it would never execute if `autoBounds` is disabled.
* Added `RenderContext.tmpPoint` alongisde `tmpBounds` - used for when `autoBounds` is disabled to transfer scaling.
* AbstractMask got new internal variables in order to keep consistent with resolution scaling. Also exactly because of AbstractMask using double-filter approach - resolution variables are non-inline properties.
* Added `Filter.resolutionScale` that will affect the rendering resolution.
* Added `Filter.useScreenResolution` that will use scene scale mode to map filter texture to screen pixels. Fixes long-standing issue of `scaleMode` downscaling filter textures for PA games. Is affected by `resolutionScale`.
* `Filter.getBounds` not receives `scale:Point`, which will contain the resolution scaling dictated by by resolutionScale/useScreenResolution. Can be modified in order to alter the resulting resolution scale.

# Unresolved issues
* Resolution scaling does not work properly if filter is attached to a Scene. 
Easy solution is to just ignore it and add a note that resolution scaling is not supported for Scene. But I'd rather fix the issue.
Another solution is to override clipBounds entirely and use slightly different math in clipping logic, in order to avoid incorrect clipping. Mainly publishing this PR currently so I can work on docs instead of keeping those changes in git stash.


ncannasse:
What is the status of this PR ? ;)

Yanrishatum:
Same as it was at a time of posting: Resolution scaling still does not properly work when attached to the Scene. If you want, I can get to fixing it faster. ;)

Yanrishatum:
@ncannasse Ready for review btw.

Yanrishatum:
Thanks to @eoy, discovered a number of issues with current implementation. Working on a fix.

Yanrishatum:
Issues are ironed out. Waiting for review. @ncannasse

h2d/RenderContext.hx
```diff
@@ -77,6 +77,13 @@ class RenderContext extends h3d.impl.RenderContext {
 	**/
 	@:dox(hide)
 	public var tmpBounds = new h2d.col.Bounds();
+	/**
+		<span class="label">Internal usage</span>
+		
+		Used to calculate filter rendering resolution.
+	**/
+	@:dox(hide)
+	public var tmpPoint = new h2d.col.Point();
```
ncannasse, line 86: please make it private without documentation and access it with @:privateAccess, it's not meant to be public but implementation detail. Alternatively, you can use a static var in Object.

h2d/filter/AbstractMask.hx
```diff
@@ -7,6 +7,9 @@ class Hide extends Filter {
 	public var input : h2d.Tile;
 	public var maskVisible : Bool;
 
+	public var inputWidth:Int;
+	public var inputHeight:Int;
```
ncannasse, line 11: Any reason to make them public ?

ncannasse:
I haven't double checked all the scaling calculus but I guess you tested it well ;)

PR #904, opened 2020-11-03
Fix Std.is deprecation warnings

Yanrishatum:
Added `hxd.impl.Api.is` which replaces `Std.is` calls due to it being deprecated in favor of `Std.isOfType`. Fixes deprecation warnings for upcoming Haxe 4.2 onwards.


Yanrishatum:
Okay, I only now noticed that it just replaced one warning to another (i.e. `Warning : Using "is" as an identifier is deprecated`). I'll do a follow up PR that will change the method name.

PR #911, opened 2020-11-18
HXSL: Added texture size method support

Yanrishatum:
Added support for `sampler.size()` and `sampler.isize()` methods that return `vec` and `ivec` respectively with the texture dimensions. 
For Sampler2D and SamplerCube it's `vec2` and Sampler2DArray is `vec3` with Z being element count.
As with `fetch`, there is no support for GLES2.0, since I have no idea how to implement hidden uniform with texture size as a substitute.

### DIFF
--- a/hxsl/Ast.hx
+++ b/hxsl/Ast.hx
@@ -207,9 +207,6 @@ enum TGlobal {
 	Texel;
 	TexelLod;
 	TextureSize;
-	TextureSizeLod;
-	ITextureSize;
-	ITextureSizeLod;
 	// ...other texture* operations
 	// constructors
 	ToInt;
@@ -246,9 +243,6 @@ enum TGlobal {
 	ChannelFetch;
 	ChannelFetchLod;
 	ChannelTextureSize;
-	ChannelTextureSizeLod;
-	IChannelTextureSize;
-	IChannelTextureSizeLod;
 	Trace;
 	// instancing
 	VertexID;
--- a/hxsl/Ast.hx
+++ b/hxsl/Ast.hx
@@ -205,7 +205,6 @@ enum TGlobal {
 	Texture;
 	TextureLod;
 	Texel;
-	TexelLod;
 	TextureSize;
 	// ...other texture* operations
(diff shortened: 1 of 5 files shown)


ncannasse:
What's the advantage of keeping ITextureSize ? Also, maybe we should merge
the *Lod variants by making lod argument optional ? That would reduce the 4
cases to 1.

Le mer. 18 nov. 2020 à 20:18, Pavel Alexandrov <notifications@github.com> a
écrit :

ncannasse:
This would also match GLSL spec that we are following.

Le mer. 18 nov. 2020 à 20:28, Nicolas Cannasse <ncannasse@gmail.com> a
écrit :

Yanrishatum:
This is not completely following GLSL spec in current implementation anyway, since GLSL only supports `ivec` version: https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/textureSize.xhtml (While HLSL uses either Float or UINT)
But HXSL implementation does not allow direct conversion from `ivec` to `vec` (i.e. `vec2(textureSize(sampler))` ), so I had to make a compromise with int/float versions. If I to follow spec completely, that would incur a major pain in the ass to use returned results in calculations, because of incompatible types. Right now it's a mix between GLSL/HLSL.

As for lod versions, sure, I'll try. On that node, I can also change TexelFetch to omit `fetchLod` version, since by GLSL spec it always requires LOD?

ncannasse:
Sure.
And I think for now we can simply forget about the int versions that does
not provide much advantage.

Le mer. 18 nov. 2020 à 20:48, Pavel Alexandrov <notifications@github.com> a
écrit :

Yanrishatum:
* Removed `TextureSizeLod`, `ITextureSize`, `ITextureSizeLod` and `TexelFetchLod`
* `.size()` now can take optional LOD argument.
* `.fetch()` not can take optional LOD argument. `.fetchLod` is an alias to `.fetch` now for backwards compatibility.

ncannasse:
Thanks !

PR #912, opened 2020-11-21
Mark base2d texture as shared

Yanrishatum:
Reasoning: Since I discovered the `@shared` qualifier, I seriously wondered why base2d does not expose its texture, since it would allow simplification of shaders by omitting double-passing of same texture during rendering of 2d elements that performs multiple texture samples. And with introduction of `.size()` support, I decided to mark it as shared.

Here's an example of shader that uses both new `.size()` feature and shared texture of Base2d to produce subpixel antialiasing (useful for PA games, and requires setting `smooth = true`):
```haxe
package h3d.shader;

class PixelartAA extends hxsl.Shader {
	
	static var SRC = {
		
		var texture : Sampler2D;
		@var var calculatedUV:Vec2;

		function subpixelAA(uv:Vec2):Vec2 {
			var size = texture.size();
			var deriv = fwidth(uv) * .5;
			var uvmin = (uv - deriv) * size; // top-left edge of the screen pixel
			var uvmax = (uv + deriv) * size; // bottom-right edge of the screen pixel
			var fmax = floor(uvmax);
			var t = (fmax - uvmin) / (uvmax - uvmin) * (fmax - floor(uvmin));
			uv = ((fmax + 0.5 - t) / size);
		}

		function __init__fragment() {
			calculatedUV = subpixelAA(calculatedUV);
		}

	}

}
```
Also question: Should I PR that shader in, or it's too specific of a shader for stock Heaps? :)

### DIFF
--- a/h3d/shader/Base2d.hx
+++ b/h3d/shader/Base2d.hx
@@ -17,7 +17,7 @@ class Base2d extends hxsl.Shader {
 
 		@global var time : Float;
 		@param var zValue : Float;
-		@param var texture : Sampler2D;
+		@shared @param var texture : Sampler2D;
 
 		var spritePosition : Vec4;
 		var absolutePosition : Vec4;
--- a/h3d/shader/Base2d.hx
+++ b/h3d/shader/Base2d.hx
@@ -26,7 +26,7 @@ class Base2d extends hxsl.Shader {
 		@var var calculatedUV : Vec2;
 
 		@const var isRelative : Bool;
-		@param var color : Vec4;
+		@shared @param var color : Vec4;
 		@param var absoluteMatrixA : Vec3;
 		@param var absoluteMatrixB : Vec3;
 		@param var filterMatrixA : Vec3;


Yanrishatum:
Marked `color` as shared as well with same justification. Use-case is fairly clear: By being able to access object tint value, it's possible to add tinting processing after `pixelColor` was calculated. In my case, I needed to bypass `pixelColor` altogether in order to make SDF outline work properly (i.e. apply tint to internal color but only tint outline faintly)

ncannasse:
I don't think that's good this way. First the names "texture" and "color" are way too generic to be shared.

Also, it's a bit against the philosophy of the framework to assume that your shader will have a texture and a color material. The idea is clearly instead to have a chain of transforms on a single pixelColor, without having to know about previous/next steps.

Being able to access the texture directly might work for some shaders but it won't in the general case since you might have other shaders "in between" doing other color transforms that will not be performed by the second fetch. So I think it's limited to some specific shaders that can only apply as "initial color transform" and for these I think passing the texture and color manually is good enough.

Yanrishatum:
I understand where this is coming from, but I do not agree that they should not be shared. Access to Base2d texture simplifies a lot, and since you need to explicitly declare sampler without qualifiers to access it (I.e. `var texture:Sampler2D`), I don't think that generic name is a problem, since `@param var texture:Sampler2D` is treated as unique and would not use the shared one.
And I absolutely do not agree that manual color/texture pass is "good enough". Take SpriteBatch that uses multiple textures for example. Manually added shader would have no idea which texture is currently attached during the drawcall, because it's automatically managed by BatchState. 
As for `color`, it's one of the `Drawable` elements, and being able to know what color tint was used without doing unnecessary legwork is a lot more robust, but generic name does pose a problem in this case, and I'm fine skipping it. But not for `texture`. 

Alternatively, I'd propose a new qualifier that lets you forcefully borrow a param from another shader. Since it has to be explicitly used - there won't be naming conflicts of shared params, and will also allow to get necessary information if user knows what he needs, and it's on his shoulders to not break the shader pipeline. I.e. something like that:
```haxe
// Should have same type/name and point at the source shader it borrow param from.
@borrow(h3d.shader.Base2d) var texture:Sampler2D;
// Alternatively (but more complex) - declare param name in meta and let user use whatever name he wants.
@borrow(h3d.shader.Base2d, texture) var sourceTex:Sampler2D;
```

HXSL is already inaccessible to most people as it is, adding yet another qualifier people has no idea about won't change much.

PR #916, opened 2020-12-14
HXSL: Implement borrow qualifier

Yanrishatum:
Implements `@borrow(path.to.Shader)` qualifier to variables.
This is essentially reverse of the `@shared` qualifier and intended only for advanced usage. (As pretty much everything HXSL, frankly.) Lets user-defined shaders to borrow uniforms from other shaders. Does not perform checks for that shader actually being attached and is completely on the conscience of the user to ensure shaders they borrow from are present. This a follow up to #912 as an alternative way to access foreign uniforms. 

<details>
<summary>Usage example</summary>

In this shader I already posted before I need to know the texture size of Base2d to perform custom sampling filter transition curve that is specifically made for pixel-art. And "just set texture to what you going to render beforehand" is not a solution.

```haxe
class PixelartAA extends hxsl.Shader {
	
	static var SRC = {
		
		@borrow(h3d.shader.Base2d) var texture : Sampler2D;
		@var var calculatedUV:Vec2;

		function subpixelAA(uv:Vec2):Vec2 {
			var size = texture.size();
			var deriv = fwidth(uv) * .5;
			var uvmin = (uv - deriv) * size; // top-left edge of the screen pixel
			var uvmax = (uv + deriv) * size; // bottom-right edge of the screen pixel
			var fmax = floor(uvmax);
			var t = (fmax - uvmin) / (uvmax - uvmin) * (fmax - floor(uvmin));
			uv = ((fmax + 0.5 - t) / size);
		}

		function __init__fragment() {
			calculatedUV = subpixelAA(calculatedUV);
		}

	}

}
```

</details>

### Reasoning
1. While I understand that shaders are designed to be fairly isolated and communicate only via shared variables, not once I found myself in a need to have access to some `@param`, and so far only solution was to make a duplicate param, and ensure that you set it to same value as the one want data from. And this is clearly not a very good solution.
2. Uniform redundancy is solved this way. If there's something you need to get from uniform that is supplied by another shader in the pipeline - you can borrow it.
3. While feature is undoubtfully dangerous since it can produce runtime compilation errors - it's a thing to many parts of shader pipeline. I think this is a responsibility of the programmer to ensure that all required shaders are in the pipeline, and if not - its their own fault.
4. As many HXSL features - this one will be obscure and unknown to many users (unless I document it, ha), and wont affect vast majority of them, but it introduces more control over the shader pipeline to people who want it. 
5. The "name is common hence can't be marked as `@shared`" is no longer applicable, since borrowing is an explicit action from the user, and it doesn't matter if name is common or not. Especially since it requires to specify which shader it borrows from. I can make it as simple as just `@borrow` but then it's not very clear what it borrows and from where. 

### Extra notes
Because of how duplicate name is implemented, common names are a big problem in general for borrow and shared qualifiers, because those checks happen only for first variable, there are edge cases that can be resolved only by accounting for all variables that were renamed due to naming collision. Such as:
```haxe
// ShaderA
@param var color:Vec4;
// ShaderB
@shared @param var color:Vec4;
// ShaderC
var color:Vec4;
// Output
@param var color:Vec4; // ShaderA
@param var color2:Vec4; // ShaderB
@local var color3:Vec4; // ShaderC
// Expected result
@param var color:Vec4; // ShaderA
@param var color2:Vec4; // ShaderB + ShaderC
```
This wont compile btw, because `color3` is never assigned. Reverse is applicable as well, with `@borrow(ShaderB) var color:Vec4;`
This is obviously because ShaderB, which shares its uniform was processed after ShaderA, which occupied the `color` ID, causing it to be renamed into `color2`, and by the time ShaderC get processed and expects the shared uniform from ShaderB - it gets the `color` that is not marked as shared, causing it to be considered unique, disregarding that `color2` _is_ shared.

### DIFF
--- a/hxsl/Ast.hx
+++ b/hxsl/Ast.hx
@@ -385,7 +385,7 @@ class Tools {
 		if ( v.qualifiers != null )
 			for( q in v.qualifiers )
 				switch (q) {
-					case Borrow(s): return StringTools.startsWith(path, s);
+					case Borrow(s): return path == s;
 					default:
 				}
 		return false;
--- a/hxsl/Ast.hx
+++ b/hxsl/Ast.hx
@@ -84,6 +84,7 @@ enum VarQualifier {
 	Ignore; // the variable is ignored in reflection (inspector)
 	PerInstance( v : Int );
 	Doc( s : String );
+	Borrow( source : String );
 }
 
 enum Prec {
@@ -377,6 +378,16 @@ class Tools {
 			for( q2 in v.qualifiers )
 				if( q2 == q )
 					return true;
+		return false;
+	}
+
+	public static function hasBorrowQualifier( v : TVar, path : String ) {
+		if ( v.qualifiers != null )
+			for( q in v.qualifiers )
+				switch (q) {
+					case Borrow(s): return StringTools.startsWith(path, s);
+					default:
+				}
 		return false;
 	}
 
--- a/hxsl/Linker.hx
+++ b/hxsl/Linker.hx
@@ -8,6 +8,7 @@ private class AllocatedVar {
 	public var merged : Array<TVar>;
 	public var kind : Null<FunctionKind>;
 	public var parent : AllocatedVar;
+	public var rootShaderName : String;
 	public var instanceIndex : Int;
 	public function new() {
 	}
@@ -68,10 +69,31 @@ class Linker {
 		return Error.t(msg, p);
 	}
 
-	function mergeVar( path : String, v : TVar, v2 : TVar, p : Position ) {
+	function mergeVar( path : String, v : TVar, v2 : TVar, p : Position, shaderName : String ) {
 		switch( v.kind ) {
 		case Global, Input, Var, Local, Output:
 			// shared vars
+		case Param if ( shaderName != null ):
+			var valid = false;
+			for ( q in v.qualifiers ) {
+				switch (q) {
+					case Borrow(source):
+						if ( StringTools.startsWith(shaderName, source) )
+							valid = true;
+						break;
+					default:
+				}
+			}
+			for ( q in v2.qualifiers ) {
+				switch (q) {
+					case Borrow(source):
+						if ( StringTools.startsWith(shaderName, source) )
+							valid = true;
+						break;
+					default:
+				}
+			}
+			if (!valid) throw "assert";
 		case Param, Function:
 			throw "assert";
 		}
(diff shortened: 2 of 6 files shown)


trethaller:
Note to @ncannasse : I didn't know `@shared` existed and wasn't able to find any usage of it in heaps, hide, Northgard, Darksburg, and newer projects. So backward-compat doesn't seem like a big issue

hxsl/Linker.hx
```diff
@@ -68,10 +69,31 @@ class Linker {
...
+			for ( q in v2.qualifiers ) {
+				switch (q) {
+					case Borrow(source):
+						if ( StringTools.startsWith(shaderName, source) )
+							valid = true;
+						break;
+					default:
+				}
+			}
+			if (!valid) throw "assert";
```
ncannasse, line 96: Why not use hasBorrowQualifer(v) || hasBorrowQualifier(v2) ?

hxsl/Linker.hx
```diff
@@ -88,19 +110,19 @@ class Linker {
 					}
 				// add a new field
 				if( ft == null )
-					fl2.push(allocVar(f1,p).v);
+					fl2.push(allocVar(f1,p, null, null, shaderName).v);
```
ncannasse, line 113: I don't like much that shaderName seems to be requested to track all vars correctly, and yet it's an optional parameter at the end of parameters. Move it either as first optional parameter and eventually make it not-optional.

hxsl/Ast.hx
```diff
@@ -380,6 +381,16 @@ class Tools {
 		return false;
 	}
 
+	public static function hasBorrowQualifier( v : TVar, path : String ) {
+		if ( v.qualifiers != null )
+			for( q in v.qualifiers )
+				switch (q) {
+					case Borrow(s): return StringTools.startsWith(path, s);
+					default:
```
ncannasse, line 389: Why startsWith ? seems like a problem if you  @borrow(MyShader) var x : Int  and you have a MyShaderB with a var x as well, you might borrow the wrong variable

ncannasse:
Apart from my review in code, i'm fine with the feature. Might require some documentation https://github.com/HeapsIO/heaps/wiki/HXSL#variable-kind

hxsl/Ast.hx
Yanrishatum: A leftover code where I attempted to let borrower to use different name compared to source variable. Will change to strict `==`.

hxsl/Linker.hx
Yanrishatum: Mainly to avoid extra method calls. Can replace with method calls thought.

ncannasse: Saving calls here is not much of an optimization ;)

hxsl/Linker.hx
Yanrishatum: I'll move it, but I can't make it non-optional, because the shader name information is not available in all cases where variables are allocated (i.e. inside functions) and borrowing supposed to work only for params anyway, hence its passed only when allocating the general variables and omitted for everything else.

Yanrishatum:
Ping. Also regarding docs: I'll update it when PR's merged, so no worries.

ncannasse:
Thanks ! merged

PR #921, opened 2021-01-07
Added highDPI flag that enables it when creating a SDL window.

jefvel:
I added a define `-D highDPI`, that will enable it when creating a SDL window.
It will then get the raw pixel size of the screen instead of window size, using `SDL_GL_GetDrawableSize`, and set the render context dimension to that when the window is resized. 

I don't know if this is the correct way to do it, but to me it seems to work quite well.

High DPI flag off:

image

High DPI flag on:
image

Fixes https://github.com/HeapsIO/heaps/issues/589

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ bin
 .vscode
 /tools/mikktspace/out
 *.exe
+.DS_store
--- a/h3d/Engine.hx
+++ b/h3d/Engine.hx
@@ -280,7 +280,13 @@ class Engine {
 		if( height < 32 ) height = 32;
 		this.width = width;
 		this.height = height;
-		if( !driver.isDisposed() ) driver.resize(width, height);
+
+		// Set the driver size to actual pixel width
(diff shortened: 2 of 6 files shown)


ncannasse:
There's several places where we're using window.width, so I think that for HDPI the window get_width/get_height should return the actual inner pixels of the window (in hdpi mode). and mouseX/Y , resize() and mouse events should also work based on that.

This way one application will not have to change anything to go from LDPI to HDPI.

jefvel:
I'm not completely sure how to handle window resizes in that mode. If high DPI is enabled, and the window size is set in pixes, it will appear a quarter of the size of a non high DPI window. 
Should this be handled manually by the developer by checking the `windowToPIxelRatio` variable, and then multiply the initial resolution by that?

ncannasse:
Yes I think it should be handled manually wrt the initial size, as in HDPI a window of 800x600 is really smaller

PR #927, opened 2021-02-12
Fix Haxe 4.2 threading issues

### LINKED ISSUE #922: Haxe 4.2 (latest), segfault on MainLoop.add
Hello! 
I tried Heaps with latest Haxe (4.2, nightly) and got segfault on start: access violation.

Looks like it comes from here:
https://github.com/HeapsIO/heaps/blob/master/hxd/System.hl.hx#L406

Because the mutex variable is now static here, it is not yet initialized before `__init__` call in `hxd/System.hl.hx`:
https://github.com/HaxeFoundation/haxe/commit/3070d07a0a16ac07e4da006f44f9cbbbc7a041da#diff-0fad0e326002cc0420a695052be93c4ebac0136f598f014e380c6a0246958feaR65

So mutex is null at this place and application will crash:
https://github.com/HaxeFoundation/haxe/commit/3070d07a0a16ac07e4da006f44f9cbbbc7a041da#diff-0fad0e326002cc0420a695052be93c4ebac0136f598f014e380c6a0246958feaR114

Yanrishatum:
Should fix #922 for good.

* I extracted `__init__` into a separate `init()` method to avoid prioritization bugs from compiler (init code of MainLoop/Timer was not executed before init code of Heaps that relies on it) and call it on App creation.
* Instead of calling `MainLoop.tick`, `Thread.events.progress()` is called, resolving the issue.
* There's an optional `heaps_unsafe_events` flag that enables usage of `__progress`. It's more optimal, as it recycles the event array, but I disabled it by default because it relies on internal implementation of `EventLoop`.
* Because `runMainLoop` is executed from the `EventLoop.loop` of the thread, it will have 2 more events scheduled to run right after it: The Driver initialization event and MainLoop tick, but because `runMainLoop` normally exists via `Sys.exit(0)` it should not be a problem.

### DIFF
--- a/hxd/App.hx
+++ b/hxd/App.hx
@@ -36,6 +36,7 @@ class App implements h3d.IDrawable {
 	var isDisposed : Bool;
 
 	public function new() {
+		hxd.System.init();
 		var engine = h3d.Engine.getCurrent();
 		if( engine != null ) {
 			this.engine = engine;
--- a/hxd/System.hl.hx
+++ b/hxd/System.hl.hx
@@ -5,6 +5,13 @@ import sdl.Cursor;
 #elseif hldx
 import dx.Cursor;
 #end
+#if !usesys
+import haxe.MainLoop;
+#end
+#if ( target.threaded && (haxe_ver >= 4.2) )
+import sys.thread.Thread;
+import sys.thread.EventLoop;
+#end
 
 enum Platform {
 	IOS;
@@ -37,6 +44,10 @@ class System {
 
 	#if !usesys
 	static var sentinel : hl.UI.Sentinel;
+	#end
+	
+	#if ( target.threaded && (haxe_ver >= 4.2) )
+	static var mainThread : Thread;
 	#end
 
 	// -- HL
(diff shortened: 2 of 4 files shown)


ncannasse:
We need to keep hl.Api.setErrorHandler in __init__ so it does catch errors in main() before new hxd.App
Also, instead of hxd.System.init(), I prefer more a `static var init = { .... ; false; }` that auto-initiate.

Yanrishatum:
I can try to extract the sentinel initialization to `__init__`, but `MainLoop.add` absolutely has to be outside of it, because otherwise it would hard-crash due to EntryPoint/MainLoop not being initialized.

ncannasse:
> I can try to extract the sentinel initialization to `__init__`, but `MainLoop.add` absolutely has to be outside of it, because otherwise it would hard-crash due to EntryPoint/MainLoop not being initialized.

Yes I get this. But maybe it will work in static var init, as we sort inits based on dependencies.

Yanrishatum:
I left a comment about it not working out after I tested it ;) Changing it to `static var init = { ... }` caused it to hard-crash. My assumption that there's some code that is being executed just after the `__init__` stuff goes trough and `MainLoop` relies on it.

ncannasse:
I've commented there https://github.com/HaxeFoundation/haxe/issues/10114

Yanrishatum:
I don't think it's worth waiting for Haxe fix, since getting Heaps operational in 4.2 is a rather pressing issue. Current fix should operate the same as pre-4.2 with exception that `MainLoop` is being added in the post-`__init__` section. Both error handler and sentinel are still initialized in the `__init__`.

qkdreyer:
As said [here](https://github.com/HeapsIO/heaps/pull/917#issuecomment-780954781), we got to release this fast because macOS CI pipelines are broken since yesterday

Yanrishatum:
@ncannasse Reminder that this is still pending and still an issue. If you don't want the `init` code, I can reduce PR to just fix the timer update handling. Also my other PRs are blocked for the time being until this is resolved.

qkdreyer:
@Yanrishatum Did you try to run that PR on Windows ? Because I've had runtime exceptions because of it I think

Yanrishatum:
@qkdreyer Yes, I obviously did, because I run from windows. Maybe elaborate the kind of exceptions you ran into?

qkdreyer:
Nevermind @Yanrishatum, I didn't correctly pulled that PR, looks good to me

PR #928, opened 2021-02-21
Add NodeJS support

qkdreyer:


ncannasse:
We don't want a null driver for nodejs as we are using the library for NWJS which uses JS backend + NodeJS API ;)

qkdreyer:
I’m not sure to understand what you’re implying here. My goal was to run a heaps game server using Node.js on an Unix OS.

ncannasse:
Your current implementation assumes that when having -lib hxnodejs (which
defines nodejs as well i think) the app will run in commandline with no
browser webgl output, which is not the case when compiling for Nwjs which
allows both HTML5 and nodejs API.

Le dim. 28 févr. 2021 à 14:33, Quentin Dreyer <notifications@github.com> a
écrit :

qkdreyer:
Hmm I see. What do you think about checking document != null instead ?

qkdreyer:
@ncannasse Can you have a look at this PR again please ?

ncannasse:
Please fix CI errors first please :) -- apart from Vulkan one (my side)

qkdreyer:
@ncannasse should be good now

ncannasse:
This is not correct... you are setting NullDriver on HLSDL, and break gamepad support in not js platforms as well...

qkdreyer:
Is it better now ?

ncannasse:
Good for me, thanks !

PR #931, opened 2021-02-25
Add support for local group/compound collider

rcstuber:
There currently is no way to create recursive local group colliders. This is crucial however to properly use `h3d.scene.Interactive`, as picking rays get transformed into the Interactive's local space, while the colliders obtained by `getCollider()` are currently operating in world-space, causing the colliders' and the object's position not to match properly. The suggested change is fully backward-compatible.

### DIFF
--- a/h3d/scene/Object.hx
+++ b/h3d/scene/Object.hx
@@ -595,18 +595,18 @@ class Object implements hxd.impl.Serializable {
 	}
 
 	/**
-		Build and return the global absolute recursive collider for the object.
+		Build and return the global or local recursive collider for the object.
 		Returns null if no collider was found or if ignoreCollide was set to true.
 	**/
-	@:final public function getCollider() : h3d.col.Collider {
+	@:final public function getCollider( local : Bool = false ) : h3d.col.Collider {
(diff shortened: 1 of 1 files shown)


ncannasse:
I don't think that will work as each object might have an offset that won't be taken into account by the group.
You should instead have your Interactive set  useParentTransform=false with a global collider.

rcstuber:
You're right, I totally missed that the relative position of child colliders won't taken into account. Do you have a suggestion on how to make this more fail-safe? It is a big trap that Interactive's transform get applied on-top of the object's colliders. In the Heaps samples you silently bypassed the issue by adding it to the scene directly.

ncannasse:
Yeah I agree it's a bit misleading. Maybe the same should use useParentTransform=false as well.
The problem is that using both local and global colliders is useful especially when you build these from other data, but they don't mix well together.

rcstuber:
Ok thanks. I will close the PR and go with the useParentTransform=false for now. I just ran into this today an scratched my head until I found the cause.

ncannasse:
Feel free to submit a sample patch

PR #932, opened 2021-02-28
Fix haxe 4.2 update loop

### LINKED ISSUE #922: Haxe 4.2 (latest), segfault on MainLoop.add

Yanrishatum:
Now that Haxe 4.2.1 came out with MainLoop fix, the simplified version of the previous fix PR is here. Supersedes previous PR.
For obvious reasons won't work on 4.2.0, and requires 4.2.1+

close #927 ; fix #922

### DIFF
--- a/hxd/System.hl.hx
+++ b/hxd/System.hl.hx
@@ -38,6 +38,9 @@ class System {
 	#if !usesys
 	static var sentinel : hl.UI.Sentinel;
 	#end
+	#if ( target.threaded && (haxe_ver >= 4.2) )
+	static var mainThread : sys.thread.Thread;
+	#end
 
 	// -- HL
 	static var currentNativeCursor : hxd.Cursor = Default;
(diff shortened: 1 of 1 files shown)


ncannasse:
What's the idea with heaps_unsafe_events ? Is it to save an array alloc each frame ?

Yanrishatum:
Basically yes. This is an optimization Thread event loop uses when you do the even loop through it. But I'm hesitant to make it default behavior, since it relies on internal methods.

PR #934, opened 2021-03-02
addTangents for HMDModel

Sechgulo:


ncannasse:
Please note that this might not work for normal maps as it's required to use the exact same method for generating the tangents that the one that was used by the normal map authoring tool. There are several conventions and the standard used in Heaps is mikktspace. You need however to have a normal map bounded to your model textures so that FBX-to-HMD conversion generates the mikktspace compatible tangents.

I'll still merge as for some usages where you don't have a normalmap, having a tangent might be useful.

PR #938, opened 2021-03-11
Fixed default Timer smoothing to comply with its description

deepnight:
This PR fixes the default `smoothing` value of `hxd.Timer`, which wasn't complying its own description: 

- 0.0 on Hashlink target
- 0.95 anywhere else


deepnight:
Actually, the very existence of this smoothing is weird: it actually causes abnormal FPS values that diverge from the real time value, and is an oooold legacy from the flash days, where FPS was inconsistent 🤔 

I'd rather really suggest to set it to 0 for modern platforms, such as HL.

ncannasse:
Agreed

deepnight:
❤️

PR #944, opened 2021-03-23
Detect touch screen in sample

zommerfelds:
Add an touch screen detection example to "Input" sample.

Hopefully this helps with visibility: https://github.com/HeapsIO/heaps/issues/943


zommerfelds:
This is just a simple change to the input sample to showcase the touch system detection function. Do you want to take it, or drop it?

PR #953, opened 2021-04-16
added followLerp and followOffset for Camera.hx

Descrout:
Added ``followLerp`` for smooth camera follow.
Added ``followOffset`` for displacing the camera follow.
Made  ``follow`` dynamic, so the camera can follow the objects other than the ``h2d.Object``. For example a ``h2d.SpriteBatch.BatchElement``.

### DIFF
--- a/h2d/Camera.hx
+++ b/h2d/Camera.hx
@@ -218,9 +218,9 @@ class Camera {
 			var fx = follow.x;
 			var fy = follow.y;
 			
-			if ( followLead != null ) {
-				fx += followLead.x;
-				fy += followLead.y;
+			if ( followOffset != null ) {
+				fx += followOffset.x;
+				fy += followOffset.y;
 			}
 
 			this.x += (fx - this.x) * followLerp;
--- a/h2d/Camera.hx
+++ b/h2d/Camera.hx
@@ -88,7 +88,7 @@ class Camera {
 	/**
 		Makes camera to follow the referenced target position.
 	**/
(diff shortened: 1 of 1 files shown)


Yanrishatum:
I would strongly advice against making follow object as dynamic.  You still can use `{ x:Float, y: Float }` typedef if you want a wildcard, but don't make it Dynamic.

Descrout:
> I would strongly advice against making follow object as dynamic. You still can use `{ x:Float, y: Float }` typedef if you want a wildcard, but don't make it Dynamic.

I used ``{ x: Float, y: Float, rotation: Float}`` at first, although this works with ``h2d.SpriteBatch.BatchElement`` , it does not work with ``h2d.Object``. Because ``x, y, rotation`` fields of ``h2d.Object`` has setters. So I get ``Inconsistent setter for field rotation : set should be default`` error. I don't know if there is another workaround.

ncannasse:
We don't want to use Dynamic in API.

Yanrishatum:
While lerping capabilities are nice to have, they are not essential. Same thing can be achieved via `sync()` of the object that controls the camera. (Well, it's up to Cannasse if he wants that in API or not - I'm not against it)
Additionally, primary idea of enabling something other than `h2d.Object` being followed is dubious. Nothing stops you from having an invisible object that controls the camera, or as mentioned above - just manually apply camera controls. It's an unique enough object to make a dedicated Object instance or system that controls it.

PR #955, opened 2021-04-18
Transition class not instantiating or running animations in order

agdt3:
Fixes issues:
- `Transition.clone()` doesn't instantiate anim1 and anim2, which fails when `Transition.update()` triggers `anim1.update()` and `anim2.update()`
- `Transition.update()` doesn't handle the issue where both `anim1` and `anim2` start on frame 0. In that case, it updates them both simultaneously, instead of waiting for `anim1` to finish.
- `SmoothTransition` doesn't have its own `clone` method and falls back on `Transition` clone method.

### DIFF
--- a/h3d/anim/SmoothTransition.hx
+++ b/h3d/anim/SmoothTransition.hx
@@ -145,4 +145,11 @@ class SmoothTransition extends Transition {
 		return rt;
 	}
 
+	override function clone(?a : Animation) : Animation {
+		if( a == null )
+			a = new SmoothTransition(anim1, anim2, speed);
+		super.clone(a);
+		var a : SmoothTransition = cast a;
+		return a;
+	}
 }
--- a/h3d/anim/Transition.hx
+++ b/h3d/anim/Transition.hx
@@ -35,7 +35,11 @@ class Transition extends Animation {
 			a = new Transition(this.name.split("(")[0], anim1, anim2);
 		super.clone(a);
 		a.anim1 = anim1.clone();
+		a.anim1.initInstance();
+		a.anim1.loop = false;
 		a.anim2 = anim2.clone();
+		a.anim2.initInstance();
+		a.anim2.loop = false;
 		return a;
 	}
 
@@ -59,8 +63,10 @@ class Transition extends Animation {
 		while( tmp > 0 )
 			tmp = anim1.update(tmp);
 		var tmp = st;
-		while( tmp > 0 )
-			tmp = anim2.update(tmp);
+		if (anim1.frame == anim1.frameCount) {
+			while( tmp > 0)
+				tmp = anim2.update(tmp);
+		}
 		return rt;
 	}
 


ncannasse:
Regarding the update() it's really meant this way : you start playing anim2 while anim1 is still playing and blend between the two.

agdt3:
Thank you for the feedback and clarification. I'll fix this PR as soon as I have some breathing room from work.

agdt3:
PR fixed. Added null check to `onAnimEnd`.

ncannasse:
Hi,
I think we should initInstance in Transition.clone only if the cloned is instanciated. Also, I'm not sure I understand why you set loop=false as it should keep all animation properties.

agdt3:
Hello, thank you for responding again.
For you first point - this is my mistake! I removed `loop` when I thought one animation was supposed to run after another, not simultaneously.

For your second point, I thought that there were a few ways to solve this problem. The problem as I understood it was primarily here(`Animation.hx:143`):
```
public function createInstance( base : h3d.scene.Object ) {
    ...
    var a = clone();
    ...
    a.initInstance();
    return a;
}
```
and here (`Animation.hx:126`):
```
function clone( ?a : Animation ) : Animation {
  if( a == null )
    a = new Animation(name, frameCount, sampling);
  a.objects = objects;
  a.speed = speed;
  a.loop = loop;
  a.pause = pause;
  a.events = events;
  a.resourcePath = resourcePath;
  return a;
}
```
and here (`Transition.hx:32`):
```
override function clone(?a : Animation) : Animation {
  var a : Transition = cast a;
  if( a == null )
    a = new Transition(this.name.split("(")[0], anim1, anim2);
  super.clone(a);
  a.anim1 = anim1.clone();
  a.anim2 = anim2.clone();
  return a;
}
```
1. `Object.playAnimation()` creates instance of `Transition` animation. (`a.createInstance(this);`)
2. The `Transition` animation clones itself .
3. When cloning itself, the `Transition` animation clones its two child animations.
4. When child animations are cloned, they drop their `isInstance` property (*this is where the problem is*).
5. The `Transition` animation instantitates itself only (but not its children).

So there are, to my mind, three solutions (maybe more):
1. Only allow `Transition` to accept instantiated `Animation`s and reject base animations in the constructor. The advantage here is that this is sort of what you want anyway from `Transition`, the disadvantage is that it might not play nice with your existing code. This also sounds like the solution you proposed, except that it fails earlier than the `clone` method. Also, this would still require solution #2.

2. Allow `Animation.clone()` to preserve the `isInstance` property, so that when `Transition` clones its child animations, their instantiated properties are replicated.
i.e.
`Animation.hx`
```
function clone( ?a : Animation ) : Animation {
  if( a == null )
    a = new Animation(name, frameCount, sampling);
  ...
  a.isInstance = isInstance;
  return a;
}
```
The advantage here is that clone should probably do this anyway. The disadvantage is that I don't know what this could break in other contexts. Maybe you never intended to clone instances!

3. Override `initInstance` in `Transition` so that it runs `initInstance` on its child `Aniamtion`s after cloning them. 
i.e.
`Transition.hx`
```
override function initInstance() {
  isInstance = true;
  anim1.initInstance();
  anim2.initInstance();
}
```
It might make sense that when you instantiate a container like `Transition` all of its children get instantiated too (and you have other `Transition` functions apply to children, like `bind`, `unbind` and `sync`), but again, maybe you didn't intend for this happen and it will wreck all your code.

4. The solution I used: instantiate the children animations in `Transition.clone()`. This leaves the smallest footprint and doesn't touch `Animation.clone()` or the `Transition` constructor... and has the chance to break the smallest amount of code.

5. The solution you proposed, which is to throw in `Transition.clone()` if the children aren't instances, will fail because if the children are instances... and you clone them... they stop being instances (see solution #2 for how to resolve this problem).

I have tested solutions (2, 3 and 4... 1 is trivial). My personal preference, if it won't break your other code, is 2 or 3, or both 2 and 3. Cloned instances remain instances instead of reverting to the base type... and instantiating a `Transition` instantiates all of its children animations as well. It also allows a base `Transition` to accept either base `Animation`s or instantiated `Animations`. Solution #2 allows `Transition` to only work with instantiated `Animation`s. #3 with either.

But this is all up to you. I am happy to implement whichever solution (or combination) you would like me to... since a lot of this has to do with backwards compatibility.

agdt3:
I've updated this PR to include my preferred solution.

PR #959, opened 2021-06-02
Lock Mouse Editing Implementation

Jed974:
Use mouseLock variable and Point Lock API to implement a new mouse movement system.
We can now call requestPointerLock to go into lock mode : the mouse cursor is hidden and cannot move, but the mouse movement is still registered.

### DIFF
--- a/hxd/Window.js.hx
+++ b/hxd/Window.js.hx
@@ -228,12 +228,15 @@ class Window {
 
 	function set_mouseLock( v : Bool ) : Bool {
 		var customCanvas = canvas != null;
-		if (v)
-			if (customCanvas) canvas.requestPointerLock();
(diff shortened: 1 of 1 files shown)


trethaller (changes requested):

hxd/Window.js.hx
```diff
@@ -255,11 +284,22 @@ class Window {
...
+		}
+		else {
+			curMouseX = e.clientX;
+			curMouseY = e.clientY;
+		}
+		
 		event(new Event(EMove, mouseX, mouseY));
 	}
 
+	function onLockChange() {
```
trethaller, line 299: This is no longer needed and can be removed

trethaller (changes requested):

PR #964, opened 2021-06-16
Added system clipboard get/set primitives and support for TextInput

Azrou:


ncannasse:
JS implementation for hxd.System would be nice too ;)

Azrou:
It doesn't seems to exist easy-to-use clipboard primitives for JS
- For node it's not avialable in native API, it need unofficial modules
- In browser it's only asynchronous and need a text input in the DOM to be used.
So I added only an empty implementation

Yanrishatum:
For browser while you still need to do some crutching to be able to read the clipboard (or to be more precise you can do that only on `paste` event or when you have `clipboard-read` permission which is usually in the `prompt` permission state). In conclusion it's impossible to have synchronous access to clipboard in browser, but it's definitely possible to implement it in async mode.

ncannasse:
Something seems wrong in the commit wrt  System.js.hx, maybe different line endings ?

PR #965, opened 2021-06-16
Disable shadows option (including screen lights and static shadows) m…

clementlandrin:
…oved into h3d.scene.Renderer.


ncannasse:
I think there's some shadows reference in the #if editor html part ;)

PR #969, opened 2021-06-18
Remove small offsets when drawing some 2D objects

zommerfelds:
See https://github.com/HeapsIO/heaps/issues/968 and https://github.com/HeapsIO/heaps/issues/776.

### DIFF
--- a/h2d/Graphics.hx
+++ b/h2d/Graphics.hx
@@ -560,13 +560,6 @@ class Graphics extends Drawable {
 		lineTo(x + w, y + h);
 		lineTo(x, y + h);
 		lineTo(x, y);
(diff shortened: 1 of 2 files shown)


zommerfelds:
Note that this problem in `SpriteBatch` only happens when `hasRotationScale = false`.

zommerfelds:
Hi all, I partially implemented what I mentioned in https://github.com/HeapsIO/heaps/issues/776#issuecomment-869611353

Do you like this approach? Please let me know. If you like it I'll implement the bevel code path and more thoroughly test it.

zommerfelds:
Hi, any feedback here would be appreciated before I forget what this was about :)

PR #978, opened 2021-07-03
Line drawing algorithms small fixes

zommerfelds:
Since https://github.com/HeapsIO/heaps/pull/969 might take longer to finish, I factored out some unrelated fixes to the drawing algorithms I found along the way. One fix is avoiding floating point issues with sin/cos to make sure the first and last point are exactly the same:

image

The second fix is that one segment in drawRoundedRect was drawn at the wrong place.


zommerfelds:
Ping. I think the changes here shouldn't be too controversial :)

zommerfelds:
Friendly ping: can we get this merged if there are no concerns?

ncannasse:
I have some issue with your patch because it changes the number of segments : when 3 segments is specified it should do exactly 3 lineTo

zommerfelds:
What shape are you drawing?

Code:
```
class App extends hxd.App {
	static function main() {
		new App();
	}

	override function init() {
		s2d.scale(10);

		final gr = new h2d.Graphics(s2d);
		gr.beginFill(0xff0000);
		gr.lineStyle(5, 0x00ff00);

		gr.drawCircle(20, 20, 10, 6);

		gr.drawRoundedRect(80, 10, 30, 25, 10, 3);

		gr.drawRoundedRect(80, 50, 30, 20, 10, 3);

		gr.drawRoundedRect(120, 10, 30, 20.01, 10, 3);

		gr.drawRoundedRect(120, 50, 30, 19.99, 10, 3);

		gr.drawEllipse(30, 60, 30, 20, 50, 3);
	}
}
```

`HeapsIO/heaps:1.9.0`:
image


`zommerfelds/heaps:fix-shape-line-drawing` as of yesterday:
image


`zommerfelds/heaps:fix-shape-line-drawing` as of my newest commit:
image

c-g-dev:
@Yanrishatum @ncannasse @TothBenoit This PR needs to be approved and merged -- the current line drawing algorithms don't work. Currently drawRoundedRect() doesn't work with even the simplest case.

PR #985, opened 2021-07-07
Update Particles.hx

apxapob:
Now it's possible to change alpha of all particles by changing alpha of parent object


ncannasse:
That will not work as we want to inherit all parents alphas so this needs to be handled in shader.

PR #990, opened 2021-07-30
Remove hxbit serialization from Samples

Speedphoenix:
Note that the Network sample still uses hxbit.


Speedphoenix:
This allow for compiling samples that use SampleApp, fixes 02c13a7

PR #999, opened 2021-09-02
Add 3D collision sample

mswieboda:
New to heaps, and the samples are great, but couldn't find anything on 3D collision detection, so I made this to help myself, and hopefully it helps others. 

I ended up figuring out from the docs and code that [Bounds collide](https://heaps.io/api/h3d/col/Bounds.html#collide) from [Object getBounds](https://heaps.io/api/h3d/scene/Object.html#getBounds) will check for collision, but not sure if that's a correct approach.

```
var collided = player.getBounds().collide(enemy.getBounds());
```

Seems to work with fine primitives, like `Cube`, but unsure if it will work for FBX models, other primitives etc, as I assume `Bounds` is like a bounding box. That's good enough for what I'm doing, so at least I thought I'd share to the community as I'm sure others are curious how to do basic 3D box collision detection if they haven't discovered it already.

If this isn't the appropriate or best pattern to check for 3D collision, please let me know. I realize making and finding examples takes time, and I appreciate yours on all your hard work on heaps. I tried digging into some listed 3D source examples, like [Cherry-jam](https://github.com/Yanrishatum/cherry-jam/) listed on [heaps.io](https://heaps.io/documentation/full-game-samples.html), and tried scourging github for a few days and couldn't find any 3D collision detection. I'm sure there's some code out there but hard to find. If there's a more correct pattern, I'm happy to learn and will change the sample.

If there's a process for contributing and PR's, couldn't find it, but would be happy to adhere to it. Lastly, looks like the samples are not synced to [heaps.io](https://heaps.io/samples/) automatically, maybe it's a manual process, or a script to do so? Might be worth to add the newer ones, happy to help if I can.

### DIFF
--- a/samples/Collision3d.hx
+++ b/samples/Collision3d.hx
@@ -1,0 +1,94 @@
+class Collision3d extends hxd.App {
+  static inline var ENEMY_COLOR = 0xcc0000;
+  static inline var PLAYER_COLOR = 0x0000cc;
+  static inline var COLLIDED_COLOR = 0xcc00cc;
+  static inline var PLAYER_SPEED = 10;
+
+  var enemy : h3d.scene.Mesh;
+  var player : h3d.scene.Mesh;
+  var text : h2d.Text;
+
+  override function init() {
+    // checkered ground
+    var worldSize = 16;
+    var halfWorldSize = Std.int(worldSize / 2);
+    var tileSize = 2;
+    var tiles = Std.int(worldSize / tileSize);
+    var plane = new h3d.prim.Cube(tileSize, tileSize, 0);
+    plane.addNormals();
+    plane.addUVs();
+
+    for (x in 0...tiles) {
+      for (y in 0...tiles) {
+        var mesh = new h3d.scene.Mesh(plane, s3d);
+        var yOddEven = x % 2 == 0 ? 1 : 0;
+        var color = y % 2 == yOddEven ? 0x408020 : 0x204010;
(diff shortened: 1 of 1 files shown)


ncannasse:
Hi , there's already some sample as part of the Interactive sample, maybe that's what you're looking for ?

mswieboda:
Hi @ncannasse, the interactive sample is great for showcasing ray casting with the mouse between 3D objects. But what about detecting when two 3D objects collide? That's what I'm trying to implement.

Unless maybe I convert an `Object` into a bunch of `Ray`s with a direction vector? Is that a normal pattern? This seems similar how the code shared in https://github.com/HeapsIO/heaps/issues/596 was accomplishing the sphere collision:

```
//myObjects is an array containing all 3 objects
for(i in myObjects)
{
	//x,y,z is the Spheres current position,  vec is the movement vector
	var ray = Ray.fromValues(x, y, z, vec.x, vec.y, vec.z);
	var checkCol = i.getCollider().rayIntersection(ray, true);
	
	if(checkCol != -1 && checkCol <= vec.length())
	{
		Debug.log("COL: " + i.type);
	}

}
```

Is that the expected way to do it, Is there a different approach?

Just looking for how to do 3D collision detecting between two `Object`s (or `Mesh`es or `Collider`s etc).

ncannasse:
Hi, we don't have general object-object collision handled right now. Some
h3d.col primitives might be able to check if one collide with the other
(like bounds/bounds or bounds/sphere) but not all cases are covered.

PR #1003, opened 2021-09-20
Fix null pointer exception on stopCapture

zommerfelds:
It is possible to destroy a scene while performing a drag. I have code like this:

```
flow.interactive.startCapture((e2) -> {
  switch (e2.kind) {
    case EMove:
      ...
    case _:
      flow.interactive.stopCapture();
  }
});
```

My callback was being called when then scene was deleted and causing an exception inside stopCapture(). This change fixes it for me.

### DIFF
--- a/h2d/Interactive.hx
+++ b/h2d/Interactive.hx
@@ -253,6 +253,7 @@ class Interactive extends Drawable implements hxd.SceneEvents.Interactive {
 		Stops current input event capture.
 	**/
 	public function stopCapture() {
+		if ( scene == null ) return;
 		scene.stopCapture();
 	}
 


ncannasse:
This will only work if the Scene also track the Interactive, detects that the Interactive has been removed and stop capturing events then. Also need to be done for both 2D and 3D.

zommerfelds:
Hi @ncannasse,

Thanks for the answer. I did some digging into your suggestion. There are three options:
1. stopCapture when the interactive is removed => requires tracking interactive / source object
2. stopCapture when a scene is removed => doesn't require tracking but not sure if this would help anyone
3. simply protect some methods with a null check => prevents the null pointer exception but is not great since a capture might be stuck for ever

Instead of calling stopCapture we could also simply set `currentDrag` in `SceneEvents` to null, basically stopping the loop but without notifying onCancel. I feel like this might be the best solution with least breaking changes. What do you think?

zommerfelds:
Just to give some background here: I noticed that my Heaps game would crash when using [Android's gesture navigation](https://support.google.com/android/answer/9079644). This is because I destroy a scene on the back button action, but the startCapture loop is still going. To fix it I had to manually call `stopCapture` on the back event. It would be nice if Heaps automatically stopped the capture when then scene is destroyed.

If there is not much interest in this, we can close this.

PR #1006, opened 2021-10-14
Add automatic smoothing support for SDF

Yanrishatum:
* `SignedDistanceField` shader now have `autoSmoothing` flag that causes `smoothing` to be ignored and its value calculated based  on `fwidth(distance)`.
* `BitmapFont.toSdfFont` now defaults to auto-smoothing.
* `Text` now automatically enables `smooth` if it wasn't explicitly set when SDF font is used.
* `HtmlText` now uses `Text.smooth` for SDF fonts.
* Improve docs a bit.

In general this is aimed to simplify out-of-the-box SDF font usage. fwidth is not a cure-all solution but should do the trick in most cases and provide decent crispiness on fonts without fiddling with smoothing value too much. I have to note that it works a bit worse on small font scales (since sampler jumps over multiple texels).

This is a somewhat of a breaking change for people who use `toSdfFont` without specifying smoothing value, but with all honesty, everyone who uses SDF fonts sets it, because default value is inadequate and should be altered from font to font and even font scale anyway.

### DIFF
--- a/h2d/Font.hx
+++ b/h2d/Font.hx
@@ -138,7 +138,7 @@ enum FontType {
 
 		@param channel The channel that serves as distance data source.
 		@param alphaCutoff The distance value that is considered to be the edge. Usually should be 0.5.
-		@param smoothing The smoothing of edge. Lower value lead to sharper edges.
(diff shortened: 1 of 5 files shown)


ncannasse:
ping @trethaller

PR #1007, opened 2021-10-18
Allow usage of 2D Bounds as a collider

Yanrishatum:
What's the title says. Bounds already satisfies Collider interface requirements.

Example use-case is using bounding box of another object as a shape for Interactive with `obj.getBounds(interactive, interBounds)`


Yanrishatum:
Why not make a `BoundsCollider` class that extends from `Bounds` just to implement the interface then? While it does make usage of Bounds as colliders a bit more cumbersome, but because `getBounds` supports passing pre-allocated Bounds instance it's manageable and not a major problem.

PR #1017, opened 2021-11-07
Added floating point color setter

onehundredfeet:
Added floating point setter for color.  Adds ease of access and precision.

### DIFF
--- a/h3d/scene/Graphics.hx
+++ b/h3d/scene/Graphics.hx
@@ -221,7 +221,7 @@ class Graphics extends Mesh {
 		setColor(color, alpha);
 	}
 
-	public function setColorF( r : Float, g : Float, b : Float, a : Float ) {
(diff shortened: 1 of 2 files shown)


h3d/scene/Graphics.hx
```diff
@@ -221,6 +221,13 @@ class Graphics extends Mesh {
 		setColor(color, alpha);
 	}
 
+	public function setColorF( r : Float, g : Float, b : Float, a : Float ) {
```
Yanrishatum, line 224:
```suggestion
	public function setColorF( r : Float, g : Float, b : Float, a : Float = 1. ) {
```

I think it would be a good idea to make alpha value optional with default value of 1 to keep in tune with other color setting methods (`setColor` / `lineStyle`).

onehundredfeet: Updated.

onehundredfeet: Anychance someone could approve and merge?

PR #1028, opened 2021-12-07
Restore hxd.fs.FileEntry.getSign

Yanrishatum:
Fix breaking API change from ea6012ab891a703c42b6d9b4a7bafa76a9766703

That method is actually being used, for example to distinct unprocessed .gif and processed .gif -> .png spritesheet with frame metadata chunk:
https://github.com/Yanrishatum/heeps/blob/56153497b9f8b6928731065229be0540cbd55810/cherry/res/GifImage.hx#L30-L32

I know it's not on purpose, but pls stop breaking file system ;)

### DIFF
--- a/hxd/fs/FileEntry.hx
+++ b/hxd/fs/FileEntry.hx
@@ -36,6 +36,14 @@ class FileEntry {
 			throw new haxe.io.Eof();
 	}
 
+	/**
+		Read first 4 bytes of the file.
+	**/
+	public function getSign() : Int {
+		var bytes = fetchBytes(0, 4);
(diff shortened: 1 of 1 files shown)


ncannasse:
Oh I thought it was no longer used in the codebase :)

Yanrishatum:
Can you mark things like that (i.e. public API) as `@:deprecated` in the future instead of outright deleting? If it's not used by anybody - nobody will ask and it can be deleted after a while, and if it's used by somebody they can either object or just use other means to achieve same behavior with a time leeway instead of having project suddenly break. Far better for backwards-compatibility.

PR #1032, opened 2021-12-16
add suggestions for args in Console

ZwodahS:
There is an existing solution at https://github.com/HeapsIO/heaps/pull/832 which does it differently from this implementation.

This implementation uses a function that return the best match instead of a list of suggestions.
Open to suggestions on how to improve this further.


ZwodahS:
I guess I should also show an example on how I use this

```
final testNames = [....];
console.addCommand("runtest", "run a specific test", [
	{
		"name": "testName",
		"t": h2d.Console.ConsoleArg.AString,
		"argSuggestions": function(tokenized: Array<String>, arg: String) {
			return utils.findClosestMatch(testNames, arg);
		}
	}
], .... 
`findClosestMatch` essentially is the same way that commands are matched `Console.hx`

PR #1046, opened 2022-02-07
Add precision for sampler2DArray

rcstuber:
Fixes issue: https://github.com/HeapsIO/heaps/issues/1035 
Shader compilation on Chrome will throw an error when using `texture.size()` function in HXSL with any `sampler2DArray` (i.e. fragmentTextures[]). Oddly the error is only thrown when a Chrome debugger is attached. I could not find any sources as to what Chrome does differently with regards to GL error checking.

### DIFF
--- a/hxsl/GlslOut.hx
+++ b/hxsl/GlslOut.hx
@@ -659,8 +659,10 @@ class GlslOut {
 		
 		if (isVertex) {
 			decl("precision highp float;");
+			decl("precision mediump sampler2DArray;");
 		} else {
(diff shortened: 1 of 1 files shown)


ncannasse:
any reason why using mediump instead of highp in vertex shader as we do for float ?

rcstuber:
No reason, just checked the specs and `highp` seems to be the default precision for vertex shaders, so I changed that now. On desktop it will use highp anyways even when mediump is specified, but on mobile it might make a difference.

PR #1052, opened 2022-03-02
Added hxd.res.Config.ignoredDirs

deepnight:
Added a Map to res Config to ignore a specific set of directories (eg. backups or other cache related things)

### DIFF
--- a/hxd/res/Config.hx
+++ b/hxd/res/Config.hx
@@ -42,11 +42,10 @@ class Config {
 
 
 	/**
-		Directories not explored by the resource scan
+		Directory names not explored by the resource scan
+		Example: `ignoredDirs = [ "backups"=>true ]`
 	**/
-	public static var ignoredDirs = [
-		"backups" => true, // LDtk backup folders
-	];
+	public static var ignoredDirs : Map<String,Bool> = [];
 
 
 	/**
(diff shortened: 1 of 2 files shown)


ncannasse:
I think these should be configured as part of `res/props.json` which are already used for hide and convert configuration. 

```json
{
  "fs.convert" : {
      ...
  },
  "fs.exclude" : ["backups", "path/to/file.png"]
}
```

deepnight:
Well, to me it feels like that there's some obscure redundancies/overlaps between Config and the JSON then. Not sure which one is actually better here 🤔

Yanrishatum:
props.json have two major issues:
1. You can't append custom configurations from libraries (for example - add converts)
2. It's extremely undocumented and obscure.

On other hand Config can be edited by external libraries, as it's macro-time.
I think for props.json to become more usable there should be a proper documentation on how it's used as well as support for props chaining, allowing for libraries to append their own configurations. (In which case usage of ldtk library could just add `backups` into FS exclude list in its own props.json)
Heaps in general have a weird disconnect that some things have to be added via props.json, but others can be done only via macro (which also poorly documented) and vice-versa.

ncannasse:
I can merge the commit but without any default directory ignored ;)

deepnight:
Thanks, sounds good! Should I make another PR?

ncannasse:
You can update this one

Le ven. 1 avr. 2022 à 12:04, Sébastien Bénard ***@***.***> a
écrit :

deepnight:
Done! :)

PR #1056, opened 2022-03-19
Made the drawTo function in objects call onAdd on scene-less objects

jefvel:
if that's not called, the filter on the object will not be bound to it.

fixes https://github.com/HeapsIO/heaps/issues/1055


trethaller:
Could there be issues with objects having onAdd called multiple times but not onRemove?

Yanrishatum:
It could be, depending if Scene does set its own `alllocated` flag during the `drawTo` or not. If `sync` is properly called then it should set the flag, and during `dispose()` call `onRemoved()`.

PR #1073, opened 2022-05-14
h2d.SpriteBatch: Removed +0.1 offsets when !hasRotationScale

NeroGM:
Here is what it does when rendering an element : (Cross cursor is at (6,6))
image


ncannasse:
The +0.1 is necessary in order to prevent some holes being displayed between tiles in a grid when scrolling with subpixel precision.

Yanrishatum:
Then it should be an opt-in/opt-out feature. Or ensured that said `0.1` is on a pixel level, not whatever scale object currently is.

PR #1079, opened 2022-06-29
Make hxd.res.DynamicText ignore ::patte(rns):: that are not alphanumeric

Speedphoenix:
Patterns containing characters that are not alphanumeric or underscores do not turn the containing text to a function anymore.

For example:
`<t>Some text ::value(hello)::</t>`
will be ignored


Speedphoenix:
Now behaves as it used to, but the r_attr regex can be customized

PR #1084, opened 2022-07-07
Fixed EOut/Eover window events for JS target

deepnight:
Issue: current `onMouseLeave()` method fires an `EReleaseOutside` event to targets, which is inconsistent:

1. the button value is always zero, no matter if the user "left" the canvas with a mouse button pressed or not.
2. the expected event should be `EOut` here.

Fix: 
1. replaced the previously not working `EReleaseOutside` with `EOut`
2. added an `onMouseEnter` callback as well to support `EOver`


ncannasse:
What about EReleaseOutside if a EPush was triggered?

PR #1085, opened 2022-07-15
(hxd.Math) A convenience function to return a random element from a given array.

marcelEuchnerMartinez:
Might be a nice method for a game framework. It's rather a convenience function.

---
- The function might also be named something like `chooseRandomElement` instead so it won't collide with `hxd.Math.random` when typing.

---
(I also did some tests on the dispersion and it seemed alright, but I'm no mathematician...)

```haxe
class Math_randomElementFromArray {
    public static function main() {
        var a = ["A","B","C","D","E"];
        Sys.println("\n Statistics test:\n"); // = "Dispersion" ?
        var hits_per_index : haxe.ds.Vector<Int> = new haxe.ds.Vector( a.length );
        for( stat_test in 0...10 ){      // several tests
            for( iteration in 0...200 ){ // each test
                var i = _randomArrayIndex( a );
                //trace( 'i=$i: ${a[i]}' );
                hits_per_index[i] ++;
            }
            trace( hits_per_index + " average should be: " + Math.floor((stat_test+1)*200/a.length*10)/10 );
        }
        Sys.println("\n Usage test:\n");
        for( n in 0...12 )
            trace( _randomElement( a ) );
    }
    //
    // My result method for public use:
    //
    static function randomElement( array ) {
        var random_index = Math.floor( Math.random() * (array.length) );
        return array[random_index];
    }
    // private test: single parts of the method
    static function _randomElement<T>( array:Array<T> ) : T {
        var random_index = _randomArrayIndex( array );
        return array[random_index];
    }
    static function _randomArrayIndex( array ) : Int {
        return Math.floor( Math.random() * (array.length) );
    }
}
```


rcstuber:
The `shuffle` function seems already out-of-place to me, since it's not really a pure math thing. Maybe you can just keep a lib of static extensions for such convenience functions. Maybe adding things like picking a random number in a range would be more fitting for `hxd.Math`. Just my thoughts...

marcelEuchnerMartinez:
Hm. I understand. That would be an understandable reason against it.

PR #1086, opened 2022-07-19
Adding a new tutorials section. Currently containing a 2D tutorial.

marcelEuchnerMartinez:
I've been working on a tutorial to get started with Heaps and so that beginners can get started more easily and learn at least the 2D basics of Heaps.

It would be great if something like a tutorials section could be an official part of the HeapsIO project on github, in case it doesn't fit in into the `HeapsIO/heaps` project then maybe it could be something like `HeapsIO/tutorials` ?

I think it makes a huge difference for a beginner to see a whole working project setup, compared to a single code sample file.

(I haven't yet added the articles for the wiki.)

Preview of the project:
preview_oop


marcelEuchnerMartinez:
Okay, then I put it into the `techTests` repo instead...

ncannasse:
I wanted to have a collection of heaps tutorials / samples, for instance I could also add a few of my games (such as https://github.com/ncannasse/ld28) . The main problem is maintaining these samples so they compile with some heaps changes that are being made. But I guess we could have a single "tutorials" repo that would regroup all of these

marcelEuchnerMartinez:
When I got started with Heaps I looked into many of the provided ludum dare contributions, but it was a bit hard to learn from them, since they are already quite advanced, hence also, must say quite stylish/beautiful. That's why I made something new from scratch, single hxd.App files getting bigger step-by-step.

Btw I made many changes to the heaps/wiki, but I online by the github site itself. I guess I should go the "pull request route" instead, so you can check for them singlewise, if they are worth a consideration.

ncannasse:
I've created a tutorials repo here:
https://github.com/heapsio/tutorials/
Feel free to submit your tutorial as a submodule it can gets maintained independently ;)

PR #1089, opened 2022-08-07
add multiline support for TextInput

csclyde:
These changes let you set a "multiline" property on the TextInput. Doing so will change the behavior of "Enter" from blurring, to adding a newline character to the text. The change also properly handles clicking on different lines, text wrapping, selecting multiple lines, etc...

### DIFF
--- a/h2d/TextInput.hx
+++ b/h2d/TextInput.hx
@@ -34,6 +34,10 @@ class TextInput extends Text {
 	**/
 	public var inputWidth : Null<Int>;
 	/**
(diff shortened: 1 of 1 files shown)


Speedphoenix:
Hello!  
This crashes when dragging the text down outside the bounds with the mouse (with multiline set to false).

On the `Input` sample:

https://user-images.githubusercontent.com/22801009/187658399-7f4f9a38-2664-4165-86de-0a71b2998b56.mp4

Issue #1067, opened 2022-04-09
JS: sound glitches at the end


zommerfelds:
There is a very unpleasant sound glitch at the end of sounds in Heaps in the browser. Also, the sounds seems less clear overall compared to the original file.

Checkout the repro here: https://zommerfelds.github.io/heaps-sound-bug/

Tested on Windows on Google Chrome and Microsoft Edge and on Android in Chrome. Checked both with embedded res or PAK file. Tested on 1.9.0 and on the latest git commit.

zommerfelds:
I just tested it in the HL output as well. There it sounds fine. Strangely however, the sound can only be played once. Maybe I should file another bug. See [source code](https://github.com/zommerfelds/heaps-sound-bug/blob/main/src/App.hx).

Yanrishatum:
JS sound driver in general is pretty wonky. It does the job somewhat, however it's severely limited in what and how it can do thing.  (And WebAudio API doesn't help in that regard, as it's centered around audio workers a lot, and Heaps being a single-file-output does not allow us to use them). 
Have you reproduced issue with other sounds, or only with extremely short ones?

zommerfelds:
Length doesn't matter. But I got a new clue! I inspected the sound files where this happens: they have a sampling rate of 44100.

Seems like if I convert to 48000 the sound is fine. Checkout my link again, I uploaded a new version comparing the two.

![image](https://user-images.githubusercontent.com/1260622/162586218-2fd3358e-fc2f-42ac-8dd5-21da7efdc977.png)

zommerfelds:
Actually, scratch that. Even if I export with Audacity in the same frequency the sound is fixed. There is simply something about the original file that Heaps can't process as well as the JavaScript API.

BTW, the sound was generated with https://sfxr.me/.

If anybody is a pro at inspecting WAV files, feel free to take a look at the difference: https://github.com/zommerfelds/heaps-sound-bug/tree/main/www/res

edwood-grant:
Hey @zommerfelds,

I decide to chekc on this, and the cruacial differences between the two file sis the bit depth. 

One is 8-bit:
![imagen](https://user-images.githubusercontent.com/44873757/185731037-5ba3a66a-cf22-4d3e-b7c7-1574317d5197.png)

The other one is 16-bit:
![imagen](https://user-images.githubusercontent.com/44873757/185731050-f3acb4a0-c821-4c24-8015-9ebaf2f03210.png)

[...]

PR #1093, opened 2022-08-20
Fixed: 8-bit WAV files playing with glitches due to incorrect sample rate calculation

edwood-grant:
The WAV file specification mentions that 8-bit samples are stored as unsigned int, whereas bigger bit depths (16-bit and above) are stored in signed ints. 
    * See here: https://www.lightlink.com/tjweber/StripWav/Canon.html for details.

And in JS, the webaudio AudioBuffer specification requieres to change the sample rate to a 32-bit float value in the [-1, 1] range.
    * See here: https://developer.mozilla.org/en-US/docs/Web/API/AudioBuffer

The webaudio Driver.hx file assumes that the forma is in the [-128, 128] range and divides by 255. This works in 16-bit files, but because of the specification, its actually stored in [0, 255], with 128 being the midpoint, dividng by 255 will cause glitches in the sound as its being calculated incorrectly. 

Changing the way the sample is calculated for JS by substracting and dividing by 128 will set the range correctly and avoid sound glitches.

This PR fixes the #1067 issue and its reproduction project that containt two WAV files, one in 8-bit and the other in 16-bit, and on whcih the 8-bit WAV presetn glitches in its sound.

### DIFF
--- a/hxd/snd/webaudio/Driver.hx
+++ b/hxd/snd/webaudio/Driver.hx
@@ -130,7 +130,7 @@ class Driver implements hxd.snd.Driver {
 				if (channelCount == 1) {
 					var chn = buffer.inst.getChannelData(0);
 					for ( i in 0...sampleCount ) {
-						chn[i] = ui8[i] / 0xff;
+						chn[i] = (ui8[i] - 0x80) / 0x80;
 					}
 				} else {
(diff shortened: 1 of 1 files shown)


ncannasse:
Thanks

PR #1095, opened 2022-08-20
Update CollideCheck.hx

octplane:
- Remove Dead code
- Move var setup at beginning of functions
- Replace clever but complex division by two with explicit division
- Rotate entity later to ensure collision is computed with this frame's `rrect` rotation

### DIFF
--- a/samples/CollideCheck.hx
+++ b/samples/CollideCheck.hx
@@ -22,8 +22,8 @@ class CollideCheck extends hxd.App {
 		}
 		rrect.endFill();
 
-		rrect.x = s2d.width >> 1;
-		rrect.y = s2d.height >> 1;
+		rrect.x = s2d.width / 2;
+		rrect.y = s2d.height / 2;
 		rrect.rotation = Math.PI / 3;
 
 		line = new h2d.Graphics(s2d);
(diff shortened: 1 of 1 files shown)


Yanrishatum:
> Replace clever but complex division by two with explicit division

Any reason for it? Bitshift is much better for integer division by power of twos as it's much cheaper to calculate compared to actual division, which is always a much more expensive operation, especially in case of Haxe where it does not have an integer division support and would cause a floating point division. And using `/2` in this case would cause `rrect` position to be `.5` shifted if window/scene size is not divisible by 2.
I don't see what's so complex or clever about using bitshift in such situation.

octplane:
I wanted to change that for several reasons:

- Using bitshift for integer division usually implies that you know the underlying encoding of numbers in your language. For this case, we can find (more or less quickly) that `s2d.width` is an `Int` but this could be encoded in any way in the language
- In modern languages, it's pretty rare to use bitshift instead of integer division in general (at least in the non-gaming/demoscene/time-constrained industry)
- In other simpler samples, the division is very explicit https://heaps.io/samples/base2d.html and not confusing
- This sample is about collision check and I wanted all the rest of the code to be very straightforward in order not to confuse beginners

Maybe we could use both forms and explains that simple optimization in a comment?

Yanrishatum:
Heaps only targets HL VM, JS and C, 2 of which would use integer for numbers and JS is "who the hell knows" (or rather "number" can be stored both as an integer or a float, and using bitwise operators ensures it is an integer).

In general it's a fair reasoning, and I was mainly curious on why change a piece of code that works just fine. I still think that integer division by 2s should use bitshift, because division is goddamn expensive compared to much, MUCH simpler binary shift operation, and people (especially ones using modern languages, especially JS) need to learn to write performant code. ;)

octplane:
> people (especially ones using modern languages, especially JS) need to learn to write performant code.

I wouldn't know, I let my compiler optimize for me and I never write JS :-D

PR #1096, opened 2022-08-20
Fix cancelEvents and propagateEvents.

rainyt:
propagate default is true, It can never become false

### DIFF
--- a/h2d/Interactive.hx
+++ b/h2d/Interactive.hx
@@ -162,8 +162,8 @@ class Interactive extends Object implements hxd.SceneEvents.Interactive {
 				return;
 			}
 		}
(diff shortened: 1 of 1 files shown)


rainyt:
My test environment is Mac, and the default value of propagate returned by MAC is true. Propagateevents = false will never take effect.

ncannasse:
ping @trethaller

trethaller:
@rainyt I don't understand, by default [hxd.Event.propagate](https://github.com/HeapsIO/heaps/blob/master/hxd/Event.hx#L31) is false. What do you mean by "default value of propagate returned by MAC"?

rainyt:
Sorry, I don't know why. It's normal now. The previous default value returned is t

PR #1099, opened 2022-08-27
Support Mac Command = Ctrl

rainyt:
This is the normal behavior of Mac users.


zommerfelds:
@rainyt or @ZwodahS (from #1106), is there a way to also support the JS environment? #1106 is using `sys` which seems more generall than `hl`.

rainyt:
> @rainyt or @ZwodahS (from #1106), is there a way to also support the JS environment? BTW, #1106 is using `sys` which seems more generall than `hl`, but still doesn't include JavaScript.

Can you give it a try? I'm not sure if it works.

ncannasse:
Closing in favor of #1106

PR #1101, opened 2022-08-31
Implement mouseClip and mouseMode API

### LINKED ISSUE #568: pointer/mouse lock support
I see there's some old support for flash, but would be nice to have it supported across the targets.

Yanrishatum:
Integrates newly added mouse handling APIs into Heaps.
Adds 3 new features to `hxd.Window`:
* `mouseClip` - would restrain the mouse position within the window boundaries (actually extremely important for RTS games when using multi-monitor setup, so I expect to see that being added in Dune/Wartales/Northgard ;) Nothing is more infuriating than accidentally unfocusing the game because you moved the mouse a bit too fast and instead of stopping at the screen edge you end up clicking outside of the game)
* `mouseMode` - the mouse movement input mode.
* `setCursorPos` - what it says on the tin. Allows manipulating user cursor position. Also useful for Relative mouse mode, since real mouse position is undefined when relative mouse mode is enabled due to backend differences.
* `mouseLock` is deprecated in favor of `MouseMode.AbsoluteUnbound`.

I've added the following modes:
* `Absolute` - default and classic behavior of "mouse is where hardware cursor is".
* `Relative(callback: Event->Void)` - sets the relative mouse mode, and instead of `EMove` will invoke the callback with `relX` / `relY` being relative movement of the mouse. Callback can set `propagate` flag (and not set `cancel`), in which case `curMouseX` / `curMouseY` would be set to `relX` / `relY` and `EMove` event would be sent as usual.
* `AbsoluteUnbound` - what current `mouseLock` does on JS: Will enable relative mouse mode and would allow virtual mouse position to exit the window boundaries.
* `AbsoluteWrap` - same, but instead of exiting boundaries it would wrap around.

Additionally I modified `CameraController` code so it would enter `AbsoluteUnbound` mode when rotating, allowing for a much better rotation experience.

This is a draft implementation because for now it's only DX implementation (still need to add SDL and JS + test), as well as I'd like some feedback on API itself while I still work on it, instead of having to rework big chunks of it afterwards.

- [x] Finalize public API
- [x] DirectX implementation
- [x] SDL implementation
- [x] JS implementation

Things I see issues with:
* Due to backends inconsistencies with each other, when entering relative mouse mode, the hardware cursor position becomes undefined, and when exiting the relative mode - cursor position could be anywhere. So, we need to either have an explicit warning in documentation about hardware mouse position being undefined or consistently "freeze" it in place when entering relative mode. However there's a problem with `AbsoluteUnbound` and `AbsoluteWrap` as those only move virtual mouse position. Do we restore position at the time we entered relative mode, or to current virtual position? What to do when we exit while mouse position is outside window area?
* Pressing Escape on JS would force-unlock the mouse. It has to be handled somehow. Either by having callback handle the cancel exit event or adding `onCancel` callback, in which case `AbsoluteUnbound` and `AbsoluteWrap` also should get `onCancel` argument. Or by automatically re-locking it on user input, in which case we need to decide if input that locks the mouse should be discarded or sent to Heaps regardless, and also how to handle mouse events while mouse is not locked. (discard or pass trough?)
* Current implementation does not properly handle mouse movement during MouseUp/MouseDown events, I plan to fix that.

For automation:
Close #568

### DIFF
--- a/h3d/scene/CameraController.hx
+++ b/h3d/scene/CameraController.hx
@@ -156,14 +156,23 @@ class CameraController extends h3d.scene.Object {
 			else
 				zoom(e.wheelDelta);
 		case EPush:
-			@:privateAccess scene.events.startCapture(onEvent, function() pushing = -1, e.touchId);
+			@:privateAccess scene.events.startCapture(onEvent, function() {
+				pushing = -1;
+				var wnd = hxd.Window.getInstance();
+				wnd.mouseMode = Absolute;
+				wnd.setCursorPos(Std.int(pushStartX), Std.int(pushStartY));
+			}, e.touchId);
 			pushing = e.button;
 			pushTime = haxe.Timer.stamp();
 			pushStartX = pushX = e.relX;
 			pushStartY = pushY = e.relY;
+			hxd.Window.getInstance().mouseMode = AbsoluteUnbound;
 		case ERelease, EReleaseOutside:
 			if( pushing == e.button ) {
 				pushing = -1;
+				var wnd = hxd.Window.getInstance();
+				wnd.mouseMode = Absolute;
+				wnd.setCursorPos(Std.int(pushStartX), Std.int(pushStartY));
 				@:privateAccess scene.events.stopCapture();
 				if( e.kind == ERelease && haxe.Timer.stamp() - pushTime < 0.2 && hxd.Math.distance(e.relX - pushStartX,e.relY - pushStartY) < 5 )
 					onClick(e);
--- a/h3d/scene/CameraController.hx
+++ b/h3d/scene/CameraController.hx
@@ -166,7 +166,7 @@ class CameraController extends h3d.scene.Object {
 			pushTime = haxe.Timer.stamp();
 			pushStartX = pushX = e.relX;
 			pushStartY = pushY = e.relY;
-			hxd.Window.getInstance().mouseMode = AbsoluteUnbound;
+			hxd.Window.getInstance().mouseMode = AbsoluteUnbound(true);
 		case ERelease, EReleaseOutside:
 			if( pushing == e.button ) {
 				pushing = -1;
--- a/hxd/Window.hl.hx
+++ b/hxd/Window.hl.hx
@@ -65,6 +65,8 @@ class Window {
 	var windowHeight = 600;
 	var curMouseX = 0;
 	var curMouseY = 0;
+	var startMouseX = 0;
+	var startMouseY = 0;
 	var savedSize : { x : Int, y : Int, width : Int, height : Int };
 
 	static var CODEMAP = [for( i in 0...2048 ) i];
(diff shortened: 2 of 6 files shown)


trethaller:
I was wondering about the use case for AbsoluteWrap and that's how Blender does it for camera navigation. 
I'm guessing the reason is that when freezing the cursor in place, it can feel unintuitive to find it back at the mouse press position, when we expect it to have traveled some distance.

Yanrishatum:
> JS: could we revert to Absolute mode when escape is pressed on JS?

Then we need a way to know it left the relative mode, hence `onCancel`, but then said `onCancel` have to be added to any mode that is not `Absolute`.

Yanrishatum:
* Relative modes now have a `restorePos` flag. Takes action only when changing back to `Absolute` mode. If set, will restore the position of mouse it had when entering the relative mode. Otherwise would clip the cursor position to window boundaries (mainly for Relative and AbsoluteUnbound that went outside window area)
* Added `Window.onMouseModeChange` that is invoked when `mouseMode` is about to change. Can return another mouse mode that would override whatever value was supposed to be set. This can be used for JS in order to keep relative mode active when browser force-exits it.
* Implemented HLSDL and JS support.

Now, it's pretty much done with exception of a few JS-related unresolved questions:
1. Should mouse inputs be ignored if current mode is relative, but mouse is not captured?
2. Should we discard mouse input from user when he causes the capture to activate?

Given the situation:
We enter relative mouse mode, mouse gets captured and everything works great. Now user alt-tabs out of the window, causing capturer to cancel and enter `Absolute` mode. Since we still want to go relative, through `onMouseModeChange` we force it back into relative. Then user activates the window again, but mouse is no longer captured. By moving mouse currently he still would cause movement to pass through to Heaps and register (question 1). Then he clicks the canvas, causing the mouse capture to activate. Right now it will also trigger a EPush+ERelease events (question 2).

Alternative solution: Just make it configurable via `Window.discardPointerCaptureEvent` and `Window.ignoreOutOfCaptureEvents` (then question is what are the defaults).

Yanrishatum:
For reference behaviors of other backends:
SDL:
* Mouse inputs are ignored when not focused.
* When gaining focus via click will cause a relative movement event towards the click position, click is discarded.
* When gaining focus via alt-tab will cause a relative movement event towards window center.
DX:
* Mouse inputs are ignored when not focused.
* When gaining focus via click will not discard the click event.
* When gaining focus via alt-tab nothing on note happens.

So yeah, only consistency we can achieve is ignoring mouse inputs. I personally lean towards ignoring inputs and to discard the mouse input that causes recapture to occur.

trethaller:
Everything looks good to me, except I'm not sure AbsoluteWrap(true) makes any sense, and since AbsoluteWrap is a pretty specific use case that can be emulated, I think it's better to remove it

PR #1103, opened 2022-09-07
Use Format to load TGA files

nspitko:
Currently, the heaps TGA loader only supports 32 bit TGAs. 

I've found this problematic in a lot of my recent work, which has involved 8 and 24 bit TGAs (For example, q3map2 spits out 24 bit TGAs for lightmaps). 

The Format lib has supported tga for a while now, so it seems reasonable we can just leverage that work.

Notes:
- Format takes care of image origin for us, so the flip code was safe to remove.
- extract32 returns BGRA. I've used the format conversion code from DDS here to convert if needed. This will incur a small perf cost as it appears (at least in hldx) we explicitly request RGBA from `toTexture()`.

### DIFF
--- a/hxd/res/Image.hx
+++ b/hxd/res/Image.hx
@@ -238,7 +238,7 @@ class Image extends Resource {
 
 		case _ if( entry.extension == "tga" ):
 			inf.dataFormat = Tga;
-			inf.pixelFormat = ARGB;
+			inf.pixelFormat = BGRA;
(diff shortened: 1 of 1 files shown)


ncannasse:
I think we used the manual code so it's optimized for HL (to get good performances with large TGA), but maybe there's room to have either both implementation or backport the optimizations into format directly ?

PR #1106, opened 2022-10-10
use CMD instead of CTRL for mac for modifier key

ZwodahS:
CTRL + C -> CMD +C etc


ZwodahS:
Didn't see that there is another PR for this
https://github.com/HeapsIO/heaps/pull/1099.

ncannasse:
The fix is good but would require a more cross platform detection of Mac in hxd.System that would support JS as well.

ZwodahS:
I am not sure when I can update this. I am in the middle of another project and my heaps is behind by a lot.

I will handle this the next time I update my heaps. If anyone needs this, feel free to fix this. 

Thanks !

PR #1112, opened 2022-11-17
JS: Fix input bugs with window resize.

Yanrishatum:
Fixes an issue with mouse positioning being wrong in case window got resized but canvas didn't. 

Honestly entire mouse positioning should be rewritten and not rely on canvas bounding box arbitrarily fetched at resize events/init. Mouse input likely also will bug out in case of window scroll and canvas being not in the window-position of the bounding box.

### DIFF
--- a/hxd/Window.js.hx
+++ b/hxd/Window.js.hx
@@ -123,9 +123,9 @@ class Window {
 				checkResize();
 			});
 			observer.observe(canvas);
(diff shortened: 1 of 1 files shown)


MSGhero:
I struggled with this for a while and ultimately thought I needed to dedicate time to figure out how to make it work. But it was a bug...

PR #1121, opened 2023-01-10
Compile define -D heaps_no_error_trap to suppress trapping exceptions…

onehundredfeet:
… in heaps mainloop

During development, we often want the application to crash as opposed to keep running when it encounters an uncaught exception.

### DIFF
--- a/hxd/System.hl.hx
+++ b/hxd/System.hl.hx
@@ -139,8 +139,10 @@ class System {
 		var eventRecycle = [];
 		#end
 		while( true ) {
+			#if !heaps_no_error_trap
 			try {
 				hl.Api.setErrorHandler(reportError); // set exception trap
+			#end
 				#if ( target.threaded && (haxe_ver >= 4.2) )
 				// Due to how 4.2+ timers work, instead of MainLoop, thread events have to be updated.
 				// Unsafe events rely on internal implementation of EventLoop, but utilize the recycling feature
@@ -155,12 +157,11 @@ class System {
 				#end
 
 				if( !mainLoop() ) break;
+			#if !heaps_no_error_trap
 			} catch( e : Dynamic ) {
 				hl.Api.setErrorHandler(null);
-				#if heaps_rethrow_error
-				hl.Api.rethrow(e);
-				#end
 			}
+			#end
 			#if hot_reload
 			if( check_reload() ) onReload();
 			#end
--- a/hxd/System.hl.hx
+++ b/hxd/System.hl.hx
@@ -139,10 +139,8 @@ class System {
 		var eventRecycle = [];
 		#end
 		while( true ) {
-			#if !heaps_no_error_trap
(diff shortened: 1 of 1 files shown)


ncannasse:
What is that used for ? if you're using the debugger it won't change much, and if not it's better to have the choice to continue or exit ;)

onehundredfeet:
Unfortunately, I'm on an M1 Ultra Mac.  The debugger doesn't work.  I was using the debugger up until upgrading my machine.

In order to find some very difficult bugs, I needed it to exit immediately on encountering the exception instead of just messaging.

I need the full call stack so, throwing from the source is important rather than just re-throwing it.

ncannasse:
You can use hl.Api.rethrow to keep the stack

Le ven. 13 janv. 2023 à 00:18, Ryan Cleven ***@***.***> a
écrit :

onehundredfeet:
Using that, I could limit the change to this:
```haxe
			try {
				hl.Api.setErrorHandler(reportError); // set exception trap
				#if ( target.threaded && (haxe_ver >= 4.2) )
				// Due to how 4.2+ timers work, instead of MainLoop, thread events have to be updated.
				// Unsafe events rely on internal implementation of EventLoop, but utilize the recycling feature
				// which in turn provides better optimization.
				#if heaps_unsafe_events
				@:privateAccess mainThread.events.__progress(Sys.time(), eventRecycle);
				#else
				mainThread.events.progress();
				#end
				#else
				@:privateAccess haxe.MainLoop.tick();
				#end

				if( !mainLoop() ) break;
			} catch( e : Dynamic ) {
				hl.Api.setErrorHandler(null);
			        #if heaps_rethrow_errors
			        hl.Api.rethrow(e);
			        #end
			}
			#end
```

ncannasse:
I think you can do it without any change by setting logError dynamic
function in your App.

Le ven. 13 janv. 2023 à 08:37, Ryan Cleven ***@***.***> a
écrit :

onehundredfeet:
I can't find logError in heaps.  Is it in hashlink?  did you mean reportError?

ncannasse:
Yes sorry reportError

Le ven. 13 janv. 2023 à 09:05, Ryan Cleven ***@***.***> a
écrit :

onehundredfeet:
ah. I see. as long as the call stack is preserved all the way through then it should be fine.  I'll run a check.

onehundredfeet:
this doesn't work.  It says trapped for some reason.
```haxe
		hxd.System.reportError = function(e) {
			trace('My rethrow');
			hl.Api.rethrow(e);
		}
```

ncannasse:
I think your original commit was better then because it works on all platforms and not only HL ;)

onehundredfeet:
Right :). Makes sense.  Simplification reverted.

onehundredfeet:
Thanks!

PR #1125, opened 2023-01-27
Fixed Variable ID allocation in linker

onehundredfeet:
Duplicate IDs were being created due to two different methods of UID generation.

I'll remove the comment before submitting if this change is reviewed and approved.


onehundredfeet:
I've changed this to address the 'repeatable' or 'always the same' principle that @ncannasse mentioned.

ncannasse:
I don't think that's a good fix, let's discuss further the problem in the original issue.

PR #1126, opened 2023-01-28
Fix multiply blend mode

csclyde:
It seems that in a previous update, a small mistake in the Multiple blend mode was introduced. This fixes it so that Multiply works properly.


ncannasse:
Thanks !

PR #1127, opened 2023-02-06
Implement Sampler.grad method for HXSL

Yanrishatum:
Adds support for GLSL [`textureGrad`](https://registry.khronos.org/OpenGL-Refpages/gl4/html/textureGrad.xhtml) / HLSL [`SampleGrad`](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-to-samplegrad) methods as `Sampler.grad()`.
Requested by CLYDE from Discord.

Sample shader (ported from [this](https://www.shadertoy.com/view/tttGzj) ShaderToy shader):
```haxe
class HxslGradTest extends hxsl.Shader {

	static var SRC = {
		
		@borrow(h3d.shader.Base2d) var texture: Sampler2D;

		@global var time: Float;
		var calculatedUV: Vec2;
		var pixelColor: Vec4;

		function fragment() {
			var blur = pow(.0625 - cos(calculatedUV.x * 20.0 + time) * .0625, 2.);
			var grad = vec2(blur, blur);
			pixelColor =
				texture.grad(calculatedUV + .5 * vec2( blur, blur), grad, grad) * .25 +
				texture.grad(calculatedUV + .5 * vec2( blur,-blur), grad, grad) * .25 +
				texture.grad(calculatedUV + .5 * vec2(-blur, blur), grad, grad) * .25 +
				texture.grad(calculatedUV + .5 * vec2(-blur,-blur), grad, grad) * .25;
		}
	}

}
```

### DIFF
--- a/hxsl/Ast.hx
+++ b/hxsl/Ast.hx
@@ -208,6 +208,7 @@ enum TGlobal {
 	TextureLod;
 	Texel;
 	TextureSize;
+	Grad;
 	// ...other texture* operations
 	// constructors
 	ToInt;
@@ -243,6 +244,7 @@ enum TGlobal {
 	ChannelReadLod;
 	ChannelFetch;
 	ChannelTextureSize;
+	ChannelGrad;
 	Trace;
 	// instancing
 	VertexID;
--- a/hxsl/Checker.hx
+++ b/hxsl/Checker.hx
@@ -84,6 +84,12 @@ class Checker {
 					{ args : [ { name: "tex", type: TSampler2D }, { name: "pos", type: ivec2 }, { name: "lod", type: TInt } ], ret: vec4 },
 					{ args : [ { name: "tex", type: TSampler2DArray }, { name: "pos", type: ivec3 }, { name: "lod", type: TInt } ], ret: vec4 },
 				];
+			case Grad:
+				[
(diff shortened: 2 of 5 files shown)


onehundredfeet:
Do all platforms support this instruction?  Are there any versions of GL or DX that don't?

Yanrishatum:
Yes, all platforms properly support this method. GL method is supported by GLSL/HLSL versions Heaps targets as well as WebGL.

ncannasse:
please note that we're using GLSL semantics, so while we can do   tex.grad(...) , the toplevel function should be named textureGrad(...) and not "grad".

Yanrishatum:
It's not a top-level function thought? It follows same approach as all sampler methods - `sampler.method()`, i.e. `texture()` -> `sampler.get()`, `texelFetch` -> `sampler.fetch()`, `textureSize` -> `sampler.size()` and consequently `textureGrad` -> `sampler.grad()`
If we to follow GLSL semantics, we also should remake `get`, `size` and `fetch` methods.

ncannasse:
You can use both object access and toplevel functions in HXSL, so you
simply need to rename your Global to TextureGrad

Le jeu. 16 mars 2023 à 02:36, Pavel Alexandrov ***@***.***> a
écrit :

PR #1129, opened 2023-02-09
[HXSL] Enable ability to write to arrays.

Yanrishatum:
Allows writing to array types as it is legal in both GLSL(+WGL) and HLSL. 
Feature requested by tee from Discord.

Same as #1127 I tested it on SDL, DX and JS targets, and there are no compatibility issues.

Test shader used:
```haxe

class WriteShader extends hxsl.Shader {

	static var SRC = {

		var pixelColor: Vec4;

		function fragment()
		{
			var p:Array<Vec2,3> = [vec2(0.0),vec2(0.0),vec2(0.0)];

			var i = 0;
			p[i]=vec2(1.0,1.0);

			var g = vec2(1.0,1.0);
			p[1]=g;

			pixelColor = vec4(p[0].x,p[1].y,p[2].x,1.0);
		}
	}

}
```

### DIFF
--- a/hxsl/Checker.hx
+++ b/hxsl/Checker.hx
@@ -338,6 +338,9 @@ class Checker {
 		case TSwiz(e, _):
 			checkWrite(e);
 			return;
+		case TArray(e, _):
+			checkWrite(e);
+			return;
 		default:
 		}
 		error("This expression cannot be assigned", e.p);
--- a/hxsl/Dce.hx
+++ b/hxsl/Dce.hx
@@ -149,6 +149,14 @@ class Dce {
 			writeTo.pop();
 			if( isAffected.indexOf(v) < 0 )
 				isAffected.push(v);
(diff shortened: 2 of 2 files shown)


onehundredfeet:
Are there any edge cases for this?  i.e. are there any versions of GL / DX that glitch on it? It looks pretty safe.

Yanrishatum:
> Are there any edge cases for this?
Not that I'm aware of no. Nor do I know of any glitches with array writes in either GL/DX/WGL.

ncannasse:
The patch seems too restrictive as it only supports TArray(TVar(v)) and I think that some cases are thus not correctly supported, such as arrays of arrays ?

Yanrishatum:
> such as arrays of arrays

Arrays of arrays are GL4+ core feature and requires `ARB_arrays_of_arrays` on lower version. And its support on WebGL is dubious as well, as WGL2 is ES3.0, while array of arrays extension is supported only since ES3.1, therefore there's no support of nested arrays.

ncannasse:
Seems good then, thanks !

PR #1133, opened 2023-02-20
Add extra clarification on Timer.wantedFPS

Yanrishatum:
Way, way, WAY too many people mistake `Timer.wantedFPS` value to be controlling the FPS game runs at and then complain that it does not work. And since it's such a common occurrence - adding extra clarification is required.


ncannasse:
Maybe it should then be renamed?

Yanrishatum:
Well, more apt name would `desiredTickrate`? With `wantedFPS` marked as `@:deprecated` and turned into a get/set passthrough property as to not cause breaking changes.

ncannasse:
what about `Timer.referenceFPS` ?

PR #1136, opened 2023-03-14
[HXSL] Provide better errors on display

Yanrishatum:
By using `Context.fatalError` with position data provided in macro context, instead of non-helpful "Build failure" on `class` it would highlight the actual error position and error text.

Currently if there's an error in the shader:
image
With this PR:
image

### DIFF
--- a/hxsl/Macros.hx
+++ b/hxsl/Macros.hx
@@ -436,7 +436,7 @@ class Macros {
 								fields.push(f);
 					} catch( e : Ast.Error ) {
 						fields.remove(f);
-						Context.error(e.msg, e.pos);
+						Context.fatalError(e.msg, e.pos);
 					}
 				default:
 				}


ncannasse:
We already are supposed to catch these errors and report them accurately so
I'm not sure what's the problem here ?

Le mar. 14 mars 2023 à 02:15, Pavel Alexandrov ***@***.***> a
écrit :

Yanrishatum:
Except they never were properly reported in the IDE. You had to manually open Output panel, select Haxe output and see error text here (not even Problems panel). In the code itself it was displaying "Build failure" on the `class` as shown in the screenshot, not the actual position error is at nor the text of the error.

Yanrishatum:
To be more precise it's specific to compiler diagnostics output. In Discord, @kLabz mentioned that it seems to be a bug on compiler side that it doesn't properly report the error on diagnostics side. And after some checking it seems to be correct. If `Context.error` is called on a `@:build` macro it would just report a build failure in diagnostics, instead of error the call provided. Changing it to `fatalError` fixes this behavior. It shouldn't affect anything because `Context.error` already aborts the macro call anyway.

ncannasse:
I think we should keep a normal error when building as fatal errors prevent other errors from being displayed, but we could work around the compiler bug by using  if( Context.defined("display") ) fatalError else error ? (to be tested)

Yanrishatum:
Did a quick test. It seems that `display` is not defined when compiler does diagnostics check. This is clearly a compiler-side bug, and I would just keep that PR open until bug is fixed in the compiler (so I can point it out for people). Will try to make an isolated reproduction and file an issue on Haxe repo.

Yanrishatum:
Closing as fix is implemented on compiler side.

Issue #1137, opened 2023-03-14
Creating a render zone inside another render zone can lead to issues.


tfabretti:
HaxeUI uses mask to make scroll boxes. It's that way that I stumbled upon an issue : a mask inside a mask can lead to an issue, the inner mask is not translated through the outer edge of the parent mask.

Here is the visual example :
![image](https://user-images.githubusercontent.com/18549479/225076327-ee17a461-5a6e-4b3d-b73e-67b2e06c8549.png)
The inner mask (for the inner scroll box) should go up and exit the parent mask (for the parent scroll box) but it doesn't.

After a fair bit of searching, I realized what caused the issue. Here is a bit of code in RenderContext.clipRenderZone :
```
x = Math.max( x, renderX );
y = Math.max( y, renderY );
var x2 = Math.min( x + w, renderX + renderW );
var y2 = Math.min( y + h, renderY + renderH );
```
The start position of the render zone is computed before getting its end position. In the case of the start position being outside the rendering zone, it's set to the bounds of the rendering zone. Then, the desired height is added to the start position. The problem is that the start position has been changed so the desired height doesn't start at the right place, extending the desired render zone.
With a Y offset example, the render zone should go from (x, renderY) to (x + w, y + h). Since _y_ has been changed before computing _y2_, the result is off, giving a result from (x, renderY) to (x + w, renderY + h).

By simply changing the order to have something like this :
```
var x2 = Math.min( x + w, renderX + renderW );
var y2 = Math.min( y + h, renderY + renderH );
x = Math.max( x, renderX );
y = Math.max( y, renderY );
```
I get the expected result :
![image](https://user-images.githubusercontent.com/18549479/225078793-bb4c0c5a-251e-4918-af2b-edc194006e17.png)

If it's a bug, I'm surprised it hasn't been caught before as I suspect using masks for scroll boxes is fairly standard (the thing even has scroll methods). Am I missing something there ? Or maybe no one was crazy enough to put scroll boxes inside other scroll boxes ? :D

As I don't use masks any other way, my fix may cause bugs I'm not aware of, so I opened this issue instead of directly doing a PR, in case it breaks stuff unbeknownst to me. Well, also because I don't know git and I hoped someone would do that PR for me :P

Please, let me know if I'm doing anything wrong ! <3

MSGhero:
I would make a PR even if it breaks something because it's easier for people to test the changes that way. This issue may not get much attention otherwise

tfabretti:
[...]

PR #1138, opened 2023-03-15
Fix a bug with a render zone inside a render zone.

tfabretti:
Having a render zone in a render zone can lead to an issue if the child zone is partly outside the parent zone : #1137 
To fix it, I changed the computation of the start position of the render zone after computing the end position.


tfabretti:
Hey, is there something I should do to help make this PR merged ?
Explain more ? Just wait ? :)

PR #1139, opened 2023-03-16
Add ZOrdering to SpriteBatches

keenfoong:
Currently there is no way to set the draw order of elements in SpriteBatches that I know of. I believe this discourages the use of SpriteBatches in favour of bitmaps and anims even if SpriteBatches are superior in terms of performance. There also seems to be more tools for general drawables (eg Layers, Anim) rather than for SpriteBatches. The changes I've put here is the first step to remedy that if allowed.

What it does is it adds a z-order field to `BatchElement`. `SpriteBatch` then adds/inserts it into the right spot when adding to the sprite batch. 

It also re-inserts it to the right spot on setting the zorder so it should minimize the performance penalty when this is switched on. But I'm open to additional performance improvements you might notice.

Some notes

- I've kept the `hasZOrder` flag off so as to not break backwards compatibility.
- It shouldn't interfere with existing implementations unless people have added their own zOrdering to their own inherited classes.
- Performance should be largely the same unless `hasZOrder` is switched on. 
- Even when switched on, adding elements to the end of the list should be the same as before. Though adding items to the beginning of the list will be O(n). If this is an issue we can add additional behaviour to the `before=true` flag to make it add from the start of the list instead but I think that might be over-engineering it a bit.
- I've added a `ReversedElementsIterator` so we can iterate through the elements in reverse. I needed this so I could remove items easily and to add sprites at the end of the list if the zordering is the same.
- I've also added a length to SpriteBatch to tell how many elements are in the linked list.
- Added `insertBefore` and `insertAfter` to the SpriteBatch to easily add elements before or after a certain element. I'm a bit dubious on the logic here so we might have to test this.

I've tested it on my end here and it seems fine but it's still largely untested not sure the procedure here.

Let me know if I've added too much into one merge request, or if you prefer it a bit more split up.

Here is an example below. The green sprite was added last. The bullets z-order is set to `0.0` and one of the characters z-order is set to `2.0` here so the red character is always in front of the bullets. This is all one spritebatch and 1 draw call.

https://user-images.githubusercontent.com/31301148/225494701-99b8c1dd-1d82-4071-9ae9-eae480468c5c.mp4

### DIFF
--- a/h2d/SpriteBatch.hx
+++ b/h2d/SpriteBatch.hx
@@ -4,7 +4,7 @@ import h2d.RenderContext;
 import h2d.impl.BatchDrawState;
 
 /**
-   Iterates through the BatchElement linked list.
+	Iterates through the BatchElement linked list.
 **/
 private class ElementsIterator {
 	var e : BatchElement;
@@ -22,7 +22,7 @@ private class ElementsIterator {
 }
 
 /**
-   Iterates through the BatchElement linked list in reverse (from the last element backwards to the first element).
+	Iterates through the BatchElement linked list in reverse (from the last element backwards to the first element).
 **/
 
 private class ReversedElementsIterator {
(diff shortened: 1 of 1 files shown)


MSGhero:
My opinion is that SpriteBatch could make the internal linked list a bit more accessible or provide a `reorder` or `swap` method. Z sorting doesn't seem like it should be in scope.

PR #1143, opened 2023-04-13
Added the possibility to specify and auto-call a software keyboard for all TextInput

yuxiaomao:
The idea is to add support software keyboard on some platforms (e.g. console or mobile) for all TextInput created, with a single var `h2d.TextInput.showSoftwareKeyboard` to be assigned at the start of a program.
It can then be disable in individual TextInput, if needed, by setting `useSoftwareKeyboard` to false.
When `h2d.TextInput.showSoftwareKeyboard` is set, it will be called when TextInput is focused. After the input or canceled, the TextInput will be automatically blur to allow focus it again.

### DIFF
--- a/h2d/TextInput.hx
+++ b/h2d/TextInput.hx
@@ -55,9 +55,8 @@ class TextInput extends Text {
 		When disabled, showSoftwareKeyboard will not be called.
 	**/
 	public var useSoftwareKeyboard : Bool = true;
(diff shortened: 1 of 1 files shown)


yuxiaomao:
Not sure if onCursorMove should be included in the function onTextChange, that can make sens, as if we add or remove text, cursor will often move as well

PR #1152, opened 2023-06-09
Support OGG on JS, fall back to stb

MSGhero:
Allows Ogg audio files on JS with or without `stb_ogg_sound` library included. Attempts native browser handling of the file, falls back to `stb_ogg_sound` if included, and then throws a runtime error if it was not included.

If the runtime error is not preferred on Safari, we can compile-time error if an OGG is found without `stb_ogg_sound`, forcing use of the library as a fallback.

Tested on Chromebook and iOS Safari on older heaps/haxe/etc, so I'll soon do a quick check on my dev PC just to be sure.

This preserves the ability to use `stb_ogg_sound` outside of HL or JS, but I did not test that.

### DIFF
--- a/hxd/res/Config.hx
+++ b/hxd/res/Config.hx
@@ -91,9 +91,6 @@ class Config {
 			ignoredExtensions.set("mp3", true);
 			#end
 		default:
-			#if !stb_ogg_sound
-			ignoredExtensions.set("ogg", true);
-			#end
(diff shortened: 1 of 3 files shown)


MSGhero:
Narrowed the scope to just native browser handling when possible, so ignoring OGGs is the same behavior as before

PR #1169, opened 2023-10-02
CacheFileBuilder: speed up ShaderCache creation by do not save to file at each step

yuxiaomao:


ncannasse:
Maybe we shouldn't save() by default then, and add manual save() calls at places where it's required ?

PR #1177, opened 2023-11-14
Fix SDL pad reconnect after disconnect event error

yuxiaomao:
This commit revert #1154 #323 .
It fix error of pad event id after disconnect and reconnect (new), and allow multiple pad to be connected to the system (what was fixed by #1154).
However, it does not fix the double init in #323 (Fixed SDL double init), but I couldn't reproduce the error in a normal workflow.

The reason behind that is the e.controller is different in GControllerAdded (device id) and others events (joystick id). The joystick id is used to reference pad in pads (instead if device id), so pads.remove in GControllerAdded  will remove the wrong pad.

### DIFF
--- a/hxd/Pad.hx
+++ b/hxd/Pad.hx
@@ -514,7 +514,7 @@ class Pad {
 		if( @:privateAccess sp.ptr == null )
 			return;
 		var p = new hxd.Pad();
-		p.index = index;
+		p.index = sp.id;
 		p.d = sp;
 		pads.set( p.index, p );
 		for( axis in 0...6 )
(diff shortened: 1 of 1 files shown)


yuxiaomao:
https://github.com/HaxeFoundation/hashlink/commit/a502906ce5d701987f3b83a709ebd35251077114
Will return joystick index as event->controller value, but
- Later in `initPad()`, `new GameController (gctrl_open)` requires the device id, as a result `initPad()` will fail if device id is not equal to joystick index (if we disconnect and reconnect a controller).
- I didn't find out a way to get device id from joystick index (however, get an created controller from joystick index is feasible)
- One possible solution seems to Open the controller in `event_loop`, and later `new GameController (gctrl_open)` will just gather the created controller. Personally I do not like the idea of manage Open and Close in different objects

PR #1186, opened 2024-01-18
Add variable to control filter's out of bounds clipping

Tenoly:


ncannasse:
We discussed another approach for the feature that required this change.

PR #1190, opened 2024-02-07
Added word-wrapping for words longer than maxWidth & Text input up & down + line start & end controls for multilined text

robShiro:
Having words longer than the specified Text.maxWidth would break the imposed constraint & make the width of the content larger than asked.

This fix splits words with that issue to add a line break.


ncannasse:
This is already implemented using `splitText`, and the nice thing is that it's overriden in HtmlText and can thus deal with complex HTML texts (including images), so it's the preferred method for text wrapping.

ncannasse:
* I'm not sure I understand why you're testing with two additional chars (e+ne)> maxWidth instead of just e ?
* I think it's too costly with current implementation as it will look back everytime we have a letter that is >maxWidth, making the algorithm O(n2) each time a word is over maxWidth. 
 
I think we should instead:
- retain the last break char position in current line (or 0)
- if at some point our current line width - last_break_char > maxWidth, then we have a "too long" word, which we can push within the current line, then keep processing normally the rest of the next on the next line

MSGhero:
2 things to consider:

1. Languages like Thai and Japanese don't break words with spaces, so putting the "too long word" on one line doesn't help them
2. I don't think `h2d.Text` is meant to comprehensively handle all languages

From a lot of my suffering in OpenFL, I think `h2d.Text` should handle text quickly while `package.NicerText` should handle text well. Which languages "quickly" is able to cover has to be limited for practical reasons.

ncannasse:
We support Chinese and Japanese in Charset.isBreakChar which is not only
for spaces. I'm not familiar with other languages but I think we could
support them if necessary.

robShiro:
I've optimized the word wrapping logic as you suggested & also added controls for Key.UP & Key.DOWN for multilined text.

robShiro:
Made some few tweaks _(Clean up, More Controls...)_

PR #1191, opened 2024-02-07
Fix S2D event collisions for ScaleMode case

RigidStudios:
Mouse event positions are currently relative to the window - they should most likely be relative to the width/height of a scene.
This has not been tested against all features, just Interactives.


Yanrishatum:
Note: This is an issue arisen from s2d not expecting `Engine` render resolution mismatching `Window` size, we discussed it in Discord and proper solution will be presented, do not merge yet.

RigidStudios:
@ncannasse This is ready for review

clementlandrin:
Hello, in case of a manual call to h3d.Engine.resize(), h3d.Engine.onResized() is not called, thus h2d.Scene.checkResize() is not called either and the mismatch will persist. 
I think the h3d.Engine.resize should be called at each h3d.Engine.onWindowResize and the check 
`if( w != width || h != height )`
should be used to prevent the call to driver.resize only. Then the dynamic onResized can be moved at the end of the resize function.

PR #1194, opened 2024-02-23
Display line number next to properties on domkit inspector

Speedphoenix:
This depends on this domkit PR:
https://github.com/HeapsIO/domkit/pull/55

image

### DIFF
--- a/h2d/domkit/Style.hx
+++ b/h2d/domkit/Style.hx
@@ -466,15 +466,17 @@ class Style extends domkit.CssStyle {
 				if (vs != null) {
 					v = vs.value;
 					var f = find(files, f -> f.name == vs.pos.file);
-					var count = countChar(f.txt, vs.pos.pmin);
-					var s = "" + count;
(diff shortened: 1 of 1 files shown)


Speedphoenix:
Last commit allows reading .map to display source transpiled files
image

Speedphoenix:
The question remains as to whether we want to keep the full path with folders or only the filename

ncannasse:
Where is located the SourceMap file and why it's not enabled by default ? :)

ncannasse:
Usually we keep a relative path to either the project root or the "/res"
file, so it can easily be turned into a full path

Speedphoenix:
For less, source maps are located in the same place as the generated style by default (style.css will have style.css.map in the same repository)   
To generate it:
- With the EasyLess vscode extension, in settings.json: `"less.compile": { "sourceMap": true, }`
- With the lessc command: `--source-map`

For the source map path I already used hxd.res.Loader.currentInstance so that it would be relative to "/res". My comment was about displaying only the file name for less sources in the inspector, since, at least on DSW, they are all always in the same folder
image

PR #1197, opened 2024-02-28
Fixed multiline text edition

robShiro:
- Cursor can now either be at the end of one line or at the start of the next one for the same cursor index
- Spaces aren't swallowed by splitRawText when doing line breaks anymore
- Only voluntary line breaks are now included in the multiline text edition logic
- Text isn't cut anymore when doing word wrap for words longer than maxWidth which are at the end of the text

### DIFF
--- a/hxsl/Flatten.hx
+++ b/hxsl/Flatten.hx
@@ -107,6 +107,31 @@ class Flatten {
 	}
 
 	function mapExpr( e : TExpr ) : TExpr {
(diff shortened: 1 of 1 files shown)


Yanrishatum:
> Spaces aren't swallowed by splitRawText when doing line breaks anymore

Are you certain that won't cause issues with text alignment when center/right alignment is used?

PR #1202, opened 2024-04-01
add window move callback

nspitko:
Heaps has several ways of doing window callbacks implemented, so I'm open to changing this call pattern; in this case we're using the same system as onClose (a dynamic function) but there's no need for a retval and we can't block the result so the calling pattern is simplified significantly.

The motivation for this feature is to allow multi-window apps to manage report their positions back to an underlying manager (in our case ImGui). This allows eg multi-window viewports to function.

### DIFF
--- a/hxd/Window.hl.hx
+++ b/hxd/Window.hl.hx
@@ -126,6 +126,9 @@ class Window {
 		return true;
 	}
 
+	public dynamic function onMove() : Void {
+	}
+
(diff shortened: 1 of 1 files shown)


ncannasse:
There's no need to check != null when it's a dynamic function ;)

nspitko:
You can null out a dynamic function, which might be how a user expects to "disable" their callback. The null check is basically free so it seemed fine to add as a safeguard, but I can remove it if you prefer.

PR #1207, opened 2024-04-12
Clone color vector in material clone function

rcstuber:
Cloning a material currently keeps a shared reference of color (Vector4). This can lead to unexpected results, when for instance calling `setColor()` on a cloned material all objects change to the same color.


rcstuber:
Thanks!

PR #1210, opened 2024-04-23
Fix bugs in multiline text input

nspitko:
This fixes some bugs in h2d.TextInput when `multiline` is set to true. These changes are isolated to `moveCursorVertically` which is gated behind `multiline`, so they should not affect the control when in the much more common single line mode.

* Fixes moving up/down a line when at the beginning of a line
* When moving up from the top line, move to the beginning of the line. Same behavior with down on the last line (moves to end of line)
* Fixes a bug when moving to a line that is hard wrapped shorter than the current position

I started making these changes due to a crash that was fixed in a recent commit, so it might be worth double checking this doesn't break other uses of this control, or collide with changes from others working on this feature. @robShiro

### DIFF
--- a/h2d/TextInput.hx
+++ b/h2d/TextInput.hx
@@ -193,13 +193,24 @@ class TextInput extends Text {
 				cursorIndex++;
 		case K.HOME:
 			if( multiline ) {
-				var currentLine = getCurrentLine();
-				cursorIndex = currentLine.startIndex;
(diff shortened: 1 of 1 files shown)


robShiro:
@ncannasse The changes are fine by me. Some tweaks were indeed needed & that won't disturb anything on my end

PR #1214, opened 2024-05-23
Adding transition handle to "visible" property

thalesalexcin:
Value will be set to the target when the transition is finished


ncannasse:
Actually I think we should be able to do that with any Bool value, not just `visible`, so this should be handled on the domkit library side ;)

PR #1219, opened 2024-06-26
Fix Benchmark compilation error on js target

tobil4sk:
Introduced in 7ac380e0cb458fe5919961be1e4b4f3e9be57116. This fixes the heaps.io website build.

This is also needed to get the sao sample working on js again.


tobil4sk:
Changes were applied in c6d1a51888409f4adc9b5218b0780718927685d5 and e50db82d1a0f9e93579b4e5fb0e46c81a59e901d

PR #1222, opened 2024-07-11
🐭 Update TiledMapLayer Object definition to include height and width.

mxbaylee:
In the Tiled output I got locally, I got an object similar to this, where height and width are inaccessible due to the type definition here.

```
{
                 "height":63.7613636363636,
                 "id":8,
                 "name":"",
                 "rotation":0,
                 "type":"",
                 "visible":true,
                 "width":15.6306818181818,
                 "x":32.1846590909091,
                 "y":48.1193181818182
                },
```

### DIFF
--- a/hxd/res/TiledMap.hx
+++ b/hxd/res/TiledMap.hx
@@ -5,7 +5,14 @@ typedef TiledMapLayer = {
 	var data : Array<Int>;
 	var name : String;
 	var opacity : Float;
-	var objects : Array<{ x: Int, y : Int, name : String, type : String }>;
+	var objects : Array<{
+		x : Float,
+		y : Float,
+		height : Float,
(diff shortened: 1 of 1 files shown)


Yanrishatum:
On contrary I would propose removing that horrid Tiled support outright with very simple reasoning:
It's based on an _extremely_ outdated version of Tiled, missing more than half of the features modern Tiled offers today. It confuses people to use it, instead of [better](https://github.com/Yanrishatum/tiledhx) [dedicated](https://github.com/Yanrishatum/haxe-format-tiled) libraries, and then stumble onto issues like that almost instantly. Community response was and always will be - do not use it.
And my PR of at least adding optional integration to said libraries with `#if` was dismissed.

PR #1236, opened 2024-09-14
h3d.col.Seg: Include z in distance check

veitikka:
h3d.col.Seg did not yet include z axis in the last case of the distance check.


TothBenoit:
Nice catch!

PR #1242, opened 2024-11-04
Round camera position on sync

xastor:
I have a 2D game where the camera continually moves around, and the fact that the camera calculates its position without rounding to a sub-pixel position leads to stuttering of objects that calculate its position relative to the camera position. 
This change fixes the issue for me, but I also think it is good default behavior.


xastor:
Sorry, I had not noticed the change had already been made.  Closing the PR.

PR #1243, opened 2024-11-04
Don't round camera position on sync

xastor:
I have a 2D game where the camera continually moves around, and the fact that the camera rounds its position leads to stuttering.
This change fixes the issue for me, but I also think it is good default behavior to have sub-pixel precision on the camera by default.  Maybe I am wrong though? Maybe making this configurable would be better?

### DIFF
--- a/h2d/Camera.hx
+++ b/h2d/Camera.hx
@@ -224,8 +224,8 @@ class Camera {
 				matC = scaleY * -sr;
 				matD = scaleY * cr;
 			}
-			absX = Math.round(-(x * matA + y * matC) + (scene.width * anchorX * viewW) + scene.width * viewX);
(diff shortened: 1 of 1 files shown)


Yanrishatum:
Iirc that rounding was added in order to avoid pixel bleeding during rendering, but it's been 4 years and I don't remember exactly. I would say gate it behind `heaps_h2d_enable_camera_rounding` / `heaps_h2d_enable_camera_rounding` define, depending on which default behavior is preferable for Shiro.

xastor:
Is this fine for merge?  We could also check for heaps_h2d_enable_camera_rounding!=false to have the default behaviour (without a define) to not round.

rcstuber:
1) Why introduce a new define if it can be done with a property of Camera?
2) Default should be what it was until now (rounded) or else bugs will pop up in legacy code

xastor:
I have made adjustments. Would this be suitable for merge?

PR #1245, opened 2024-11-15
Explicitly declare the precision for integer types 

Sechgulo:
Since 1.10.0 default have been set to `GL2` from previous `GL`: https://github.com/HeapsIO/heaps/commit/42cd06c9e67fe43863ad5a028e2b6ab753555d44#diff-cc7e0c2966fdfe280914fd349522e9c2d19ff41f0bfb5f4f23af4d4cee7b695aR11

This seem to have introduced a change were shaders are compiled using GLSL ES 3.0 compared to previous 1.0.
Not completly sure but GLSL ES 3.00 seem to require explicit precision declarations for int at least for certain GPUs/drivers, so GPUs that strictly follow the standard (a lot of Adreno GPUs) will not compile. Hoping that explicitly declaring fixes the problem.

### DIFF
--- a/hxsl/GlslOut.hx
+++ b/hxsl/GlslOut.hx
@@ -813,8 +813,10 @@ class GlslOut {
 			// no prec
 		} else if( isVertex )
 			decl("precision highp float;");
+			decl("precision highp int;");
 		else
 			decl("precision mediump float;");
+			decl("precision mediump int;");
 
 		initVars(s);
 
--- a/hxsl/GlslOut.hx
+++ b/hxsl/GlslOut.hx
@@ -811,13 +811,14 @@ class GlslOut {
 
 		if( isCompute ) {
 			// no prec
-		} else if( isVertex )
+		} else if( isVertex ) {
 			decl("precision highp float;");
 			decl("precision highp int;");
-		else
+		} else {
 			decl("precision mediump float;");
 			decl("precision mediump int;");
-
+		}
+	
 		initVars(s);
 
 		if( isCompute )


ncannasse:
ping @clementlandrin @TothBenoit

Sechgulo:
This is the error we're seeing (JS target)

```
Uncaught An error occurred compiling the shaders: 

	#version 300 es
	precision highp float;
	in vec2 position;
	in vec2 uv;
	in vec4 color;
	vec4 position_2;
	vec4 spritePosition;
	vec4 absolutePosition;
	out vec4 pixelColor_varying;
	out vec2 calculatedUV_varying;
	vec4 outputPosition;
	uniform vec4 vertexParams[7];
	
	vec3 tmp;
	
	void main(void) {
		spritePosition = vec4(position, vertexParams[0].x, 1.);
		absolutePosition.x = dot(vec3(spritePosition.xy, 1.), vertexParams[0].yzw);
		absolutePosition.y = dot(vec3(spritePosition.xy, 1.), vertexParams[2].xyz);
		absolutePosition.zw = spritePosition.zw;
		tmp = vec3(absolutePosition.xy, 1.);
		tmp = vec3(dot(tmp, vertexParams[3].xyz), dot(tmp, vertexParams[4].xyz), 1.);
		outputPosition = vec4(dot(tmp, vertexParams[5].xyz), dot(tmp, vertexParams[6].xyz), absolutePosition.zw);
		position_2 = outputPosition;
		gl_Position = position_2;
		calculatedUV_varying = uv;
		pixelColor_varying = vertexParams[1] * color;
		gl_Position.z += gl_Position.z - gl_Position.w;
	}
```

But can confirm, even after adding declaration for in precision the error still occurs.

lbergman:
Seems like the compilation failure happens on reload after context lost, but we cannot seem to find out what causes it. We have got a device it can be reproduced on now, but it doesn't happen on all GPUs of the same model, for example we have Samsung S80 with Adreno 618, and that works fine, while other phones with same GPU fails. It affects around 2% of our users, and must be due to a change after commit 6946044 14th of Feb, but since context lost does not give any information about the source of the error we are struggling to pin it down. 

Have been trying to make a minimal example that reproduces the issue without luck so far, but will open a separate issue if I manage, and ask @Sechgulo to close this PR since it does not seem related to precision.

PR #1248, opened 2024-12-03
Fix incorrect scissor box in h2d.Scene.captureBitmap

pkhead:
Currently, there is an issue with the OpenGL driver when using `h2d.Scene.captureBitmap`. When it sets the scissor bounds to the viewport bounds, `GlDriver.curTexture` incorrectly refers to the current texture rather than the texture being rendered into, causing the scissor box to potentially use an incorrect position. I will note that the DirectX driver has no problem here because it doesn't need to alter the scissor box if it's rendering to a texture, which the OpenGL driver needs to do.

I have fixed this by simply reordering some function calls in `h2d.Scene.captureBitmap` to ensure the correct framebuffer will be set when `setRenderZone` is called.

### DIFF
--- a/h2d/Scene.hx
+++ b/h2d/Scene.hx
@@ -905,11 +905,12 @@ class Scene extends Layers implements h3d.IDrawable implements hxd.SceneEvents.I
 			var tex = new h3d.mat.Texture(width, height, [Target]);
 			target = new Tile(tex,0, 0, width, height);
 		}
+		
+		var tex = target.getTexture();
 		engine.begin();
+		engine.pushTarget(tex);
 		engine.setRenderZone(Std.int(target.x), Std.int(target.y), hxd.Math.ceil(target.width), hxd.Math.ceil(target.height));
 
-		var tex = target.getTexture();
-		engine.pushTarget(tex);
 		var ow = width, oh = height, ova = viewportA, ovd = viewportD, ovx = viewportX, ovy = viewportY;
 		width = tex.width;
 		height = tex.height;


ncannasse:
Maybe this should be fixed at the driver level @TothBenoit ;)

pkhead:
imho I don't think its a driver issue, really. just in the captureBitmap function, setRenderZone was called when a framebuffer change hadn't yet been flushed to the driver. plus, OpenGL framebuffers are upside-down, whereas it seems DirectX framebuffers arent, so it is necessary to track the current framebuffer to see if the scissor box should be flipped; I also did the same thing when I wrote an OpenGL renderer for a project. But, I obviously am not the person who wrote all of this code, so I may be wrong.

Yanrishatum:
I have to agree with pkhead in this case, it's clearly not driver-level issue and order of instructions issue. It makes no sense to set clip rect on one target, then push another target and expect the clip rect to still be the same. This PR just corrects the ordering (push first, set clip rect second) of instructions.

TothBenoit:
At API level, scissors rect can be set independently of the render target, soI don't think it should matter here.
However a dependency between the two was introduced for a fix almost 10 years ago. The commit doesn't give any details about the fixed bug, so the change is a bit unclear to me.
 
It is good to note that DirectX 12 drivers will also have a different behaviour here, as we reset the render zone when setting a new target.

The change proposed here seems like a good workaround that won't need to be reverted after a potential driver fix. I think we can safely merge it. @ncannasse

PR #1254, opened 2025-01-07
Ktx2 compressed textures support

lbergman:
Replaces https://github.com/HeapsIO/heaps/pull/710

Support for parsing Ktx2 files with ETC or UASTC encoded textures, transcoding to BPTC, DXT, ETC2 or ASTC.

WIP since we are currently not using the res system and doing our own loading, so is not integrated with res system and only supports js target.

However we are working on a new framework where we would like to make use of res system and pak loading. I tried to implement async loading for compressed textures using the AsyncLoading flag, but could not make it work. 
Will revisit this soon and try to solve the res integration, but leaving this up here for now if anyone has ideas or is interested in getting this working for native.


hxd/Pixels.hx
```diff
@@ -536,8 +538,33 @@ class Pixels {
 
 	public static function calcDataSize( width : Int, height : Int, format : PixelFormat ) {
 		return switch( format ) {
-		case S3TC(_):
-			(((height + 3) >> 2) << 2) * calcStride(width, format);
+		case S3TC(n):
+			var w = ((width + 3) >> 2) << 2; // Round up width to next multiple of 4
+			var h = ((height + 3) >> 2) << 2; // Round up height to next multiple of 4
+			var blocks = (w >> 2) * (h >> 2); // Total number of blocks
```
ncannasse, line 544: This could be simplified without  <<2  and blocks = w*h  (same for other formats)

hxd/net/BinaryLoader.hx
```diff
@@ -32,7 +32,7 @@ class BinaryLoader {
 				onError(xhr.statusText);
 				return;
 			}
-			onLoaded(haxe.io.Bytes.ofData(xhr.response));
+			onLoaded(raw ? xhr.response : haxe.io.Bytes.ofData(xhr.response));
```
ncannasse, line 35: This seems unrelated to the PR. Please submit separately

hxd/res/Image.hx
```diff
@@ -668,8 +689,8 @@ class Image extends Resource {
 	}
 
 	public function toTile():h2d.Tile {
-		getInfo();
-		return h2d.Tile.fromTexture(toTexture()).sub(0, 0, inf.width, inf.height);
+		final tex = toTexture();
+		return h2d.Tile.fromTexture(tex).sub(0, 0, tex.width, tex.height);
```
ncannasse, line 693: This seems unrelated to the PR. Please submit separately

ncannasse:
Also I'm a bit curious whereas the support for it is widely spread enough

hxd/net/BinaryLoader.hx
lbergman: This is used by Ktx2Decoder.initTranscoder when loading the wasm module.

lbergman:
All resolved. 

If you mean support on devices, it will transcode to either BPTC, DXT, ETC2 or ASTC, which should cover pretty much all desktop and mobile. There is a CLI transcoder that could be used instead of the wasm module to enable support on native targets.

I removed support for PVRTC since that is only needed for iOS devices with pre A7 chip, so older than 2013, and could not even find a device to test on, and it has limitations.

lbergman:
Made a new PR from a fresh fork to get rid of the messy history. https://github.com/HeapsIO/heaps/pull/1260

PR #1256, opened 2025-01-15
Respect hscriptPos flag for shader live reload

### LINKED ISSUE #1261: SharedShader sys access error
Getting an error after updating to latest master:
`/hxsl/SharedShader.hx:349: characters 8-15 : Accessing this field requires a system platform (php,neko,cpp,etc.)`
We are using hscript but targeting js. 
Should I make a PR to put the Sys statement in `#if sys`?

Yanrishatum:
Live reload did not respect lack of `hscriptPos` flag, causing compilation error, as it tried to call a function on a type that didn't have it.

fix #1261


yuxiaomao:
Seems good to me x) Thank you!

PR #1258, opened 2025-01-15
Support shader_debug_dump flag on non-sys JS target.

Yanrishatum:
Dumps shaders into a `window.shaders` variable when `shader_debug_dump` is set and compiled under HTML5 without `sys`. Allows debugging shaders under HTML5 when sys target is not available.

Courtesy of today discord fun times of having to extract shader dumps from dev who only target HTML5 and is too deep in target-specific hooks to compile to HL.


ncannasse:
Thanks !

PR #1260, opened 2025-01-16
Support ktx2 compressed textures

lbergman:
Reintroduced the changes done in https://github.com/HeapsIO/heaps/pull/1254 using a clean fork since history was messy.

Support for parsing Ktx2 files with ETC or UASTC encoded textures, transcoding to BPTC, DXT, ETC2 or ASTC.

WIP since we are currently not using the res system and doing our own loading, so is not integrated with res system and only supports js target.

However we are working on a new framework where we would like to make use of res system and pak loading. I tried to implement async loading for compressed textures using the AsyncLoading flag, but could not make it work.
Will revisit this soon and try to solve the res integration, but leaving this up here for now if anyone has ideas or is interested in getting this working for native.

### DIFF
--- a/h3d/impl/GlDriver.hx
+++ b/h3d/impl/GlDriver.hx
@@ -87,6 +87,15 @@ class GlDriver extends Driver {
 	static var UID = 0;
 	public var gl : GL;
 	public static var ALLOW_WEBGL2 = true;
+
+	public var textureSupport:{
+		astc:Bool,
+		astcHDR:Bool,
+		etc1:Bool,
+		etc2:Bool,
+		dxt:Bool,
+		bptc:Bool,
+	};
 	#end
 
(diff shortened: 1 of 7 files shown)


lbergman:
Updated to work with res system, but still only supported on js target.

Let me know if you are interested in merging this....if so I have a couple of questions before finalizing PR and adding docs.

1) For the ktx binaries we use a docker image with them installed, and building them is a bit messy since it needs to update cmake. Should one simply add instructions how to install ktx tools, or should I add binaries for different archs in tools?
2) I have not implemented support for native. Since there is some logic for S3TC for native targets already it is more risky with regards to regressions, and I don't have any native projects to test on. If there is an interest in getting support for native, I can help setting up the transcoder and do a naive implementation if someone is able to test on native games using S3TC.

PR #1263, opened 2025-01-17
Frag depth

lusius:
GLSL Has a writeable global for writing directly to the depthbuffer.

SV_Depth seems to be the same in HLSL.

### DIFF
--- a/hxsl/HlslOut.hx
+++ b/hxsl/HlslOut.hx
@@ -77,7 +77,6 @@ class HlslOut {
 		m.set(BVec3, "bool3");
 		m.set(BVec4, "bool4");
 		m.set(FragCoord,"_in.__pos__");
-		m.set(FragDepth,"SV_Depth");
 		m.set(FloatBitsToInt, "asint");
 		m.set(FloatBitsToUint, "asuint");
 		m.set(IntBitsToFloat, "asfloat");


hxsl/Checker.hx
```diff
@@ -418,7 +419,13 @@ class Checker {
...
 			checkWrite(e);
 			return;
-		default:
+		case TGlobal(g):
+			switch(g) {
+				case FragDepth:
+					return;
+				default:
+			}
+			default:
```
ncannasse, line 428: please simply use case TGlobal(FragDepth)

hxsl/Checker.hx
```diff
@@ -799,12 +806,11 @@ class Checker {
...
 				}
-				if( v.type == null ) error("Type required for variable declaration", e.pos);				
-				if( isImport && v.kind == Param )
-					continue;
-
+				if( v.type == null ) error("Type required for variable declaration", e.pos);
 				if( vars.exists(v.name) ) error("Duplicate var decl '" + v.name + "'", e.pos);
 				var v = makeVar(v, e.pos);
+				if( isImport && v.kind == Param )
+					continue;
```
ncannasse, line 813: You're reverting some changes here ?

hxsl/HlslOut.hx
```diff
@@ -458,7 +459,7 @@ class HlslOut {
 		case TCall({ e : TGlobal(g = (Texel)) }, args):
 			addValue(args[0], tabs);
 			add(".Load(");
-			switch( args[0].t ) {
+			switch( args[1].t ) {
```
ncannasse, line 462: Same here ?

ncannasse:
The PR needs several fixes.
Ping @TothBenoit for other comments ?

hxsl/HlslOut.hx
lusius: Yeah sorry that seems to be some error autocorrect i don't know about. Fixing

hxsl/Checker.hx
lusius: I think I have gone against an old branch because that change doesn't seem to be against mine. Fixing this.

hxsl/HlslOut.hx
lusius: Should be resolved now. Apologize for that

hxsl/Checker.hx
lusius: done

TothBenoit:
I quickly tested the changes with this :
`class CustomDepth extends hxsl.Shader {
	static var SRC = {
		function fragment() {
			fragDepth = 0.5;
		}
	}
}`
The line is being removed during linker. hasSideEffect(e) in Ast.hx should return true for FragDepth to avoid this.

Also it would be nice to either forbid reading to fragDepth in the checker, since SV_Depth is write-only or create an intermediate value that can be read and rewritten multiple times that will write to fragDepth at the end of the shader.

lusius:
Sure thing, I'll go about fixing that.

lusius:
It seems that the variable gets opted out before getting into hasSideEffects, any clues on where to continue? @TothBenoit

lusius:
It also seems to be opted in if i import BaseMesh shader. This is weird.

lusius:
@TothBenoit I've added a reference for hasSideEffects and had to add a force inclusion in the linker for this to work. But I don't know if this would be the correct way. Also shader doesn't like when I assign a variable instead to fragDepth

class CustomDepth extends hxsl.Shader { static var SRC = { function fragment() { var x = 0.5; fragDepth = x; } } }

Crashes in compiler.

I don't know how to proceed from here.

ncannasse:
Look at the way discard is implemented.

Le mar. 21 janv. 2025, 14:06, Rasmus Brinck ***@***.***> a
écrit :

lusius:
@ncannasse I did that and it seems to work fine other than it for some reason can't be assigned a variable.

lusius:
I have tried to add a readdependency but assigned variable still gets filtered out before (i think) dce. I'm trying to get a dependency going marking my variable "x" in testshader above as dependent but so far I have failed.

Do you have any ideas on how to solve this? Marking PR as draft until this is done.

lusius:
Ping @TothBenoit , do you have time to help me out here? Would be much appreciated!

TothBenoit:
Hello,
Your variable is being removed by the DCE. You need to make sure that it is marked as used during the check method of the DCE.
You can add this case to ensure this:
```
case TBinop(OpAssign, { e : TGlobal(FragDepth) }, e2 ):
	writeTo.push(null, 15)
	check(e2, writeTo, isAffected);
	writeTo.pop();
```
This will force the DCE to keep expression e2.

However, I just realised that we should only keep the last assignment to fragDepth.
In this code, only the last 2 expressions should be kept
```
var x = 0.5;
fragDepth = x;
var y = 0.5;
fragDepth = y;
```
This could be done by marking  only the last assignment to fragDepth as having a side effect in the checker, and making sure that the fragDepth assignment has a side effect in the DCE.

lusius:
@TothBenoit @ncannasse 

I have updated now ackording to the suggestions. To be able to distinguish between fragdepth writes I had to create a common variable id for it. 

``       function fragment() {
            var b = 0.3;
            fragDepth = b;
            var x = 0.5;
            fragDepth += x;
            var y = 0.5;
            fragDepth = mix(0,y,0.5);
            var a = 0.1;
            fragDepth += a;
        } 
        ``
        
        The DCE seems to handle this correct:
        ``
     y = 0.5;
    gl_FragDepth = mix(0., y, 0.5);
    a = 0.1;
    gl_FragDepth += a;
        ``

TothBenoit:
Just tested it, works fine with GL.

However, it crashes on DirectX 11 and 12. SV_Depth is not a variable, it is a semantic name :
https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-semantics

You need to declare a new output for the fragment. See FragCoord in HlslOut.hx for a similar case.

lusius:
@TothBenoit Thank you for your reply. I have removed the Hlsl-lines due to no knowledge in DX and I have no stable environment to test this properly. Thank you for you invaluable knowledge.

I'm letting the next person do the DX implementation of this.

Are we fine to merge this @ncannasse ?

PR #1265, opened 2025-01-19
Add global shader list support to s2d

Yanrishatum:
Introduces ability to add global shaders to 2D rendering pipeline. 

Rationale: Allows users to add shaders to multiple rendered objects without having to add same shader to all of them, or apply a shader that have to be always present on all 2D objects. This reduces the amount of tedious shader management. 
Use-case examples:
* Custom global filtering. In my case bilinear filtering that mimics nearest neighbor but offers reasonably antialiased transition between texels on rotated sprites.
* (From discord) Adding custom pass outputs and shader that outputs to it.
* Shader that is applied to all object children, simplifying the shader management code, as instead of having to add shader to everything in the object tree and tracking that, it's possible to just add shader as global one during root object rendering, and removing it afterwards.

Adds the following API:
* `RenderContext.addGlobalShader(shader)`
* `RenderContext.removeGlobalShader(shader)`

A few notes:
* Is there any reason for `currentShaders` to even exist? It seems fairly redundant, as `initShaders` only ever called with `baseShaderList`, and thus it can never be anything other than `baseShaderList`. Some legacy code leftovers?
* Shaders are added without sort, as they are sorted anyway by `Cache.compileRuntimeShader`
* It probably would be better to rename `baseShaderList` to `globalShaderList`.

### DIFF
--- a/h2d/RenderContext.hx
+++ b/h2d/RenderContext.hx
@@ -93,8 +93,11 @@ class RenderContext extends h3d.impl.RenderContext {
 	var compiledShader : hxsl.RuntimeShader;
 	var fixedBuffer : h3d.Buffer;
 	var pass : h3d.mat.Pass;
+	// ...Tail points at an end of global shader list, and shaders afterwards belong to specific objects.
 	var currentShaders : hxsl.ShaderList;
+	var currentShadersTail : hxsl.ShaderList;
 	var baseShaderList : hxsl.ShaderList;
+	var baseShaderListTail : hxsl.ShaderList;
 	var needInitShaders : Bool;
 	var currentObj : Drawable;
 	var stride : Int;
@@ -145,6 +148,7 @@ class RenderContext extends h3d.impl.RenderContext {
 		baseShader.setPriority(100);
 		baseShader.zValue = 0.;
 		baseShaderList = new hxsl.ShaderList(baseShader);
+		baseShaderListTail = baseShaderList;
 		targetsStack = [];
 		targetsStackIndex = 0;
 		cameraStack = [];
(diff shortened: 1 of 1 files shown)


h2d/RenderContext.hx
```diff
@@ -305,6 +310,50 @@ class RenderContext extends h3d.impl.RenderContext {
...
+
+		@returns `true` if shader was removed, `false` otherwise.
+	**/
+	public function removeGlobalShader( shader : hxsl.Shader ) {
+		if ( shader == baseShader ) throw "Cannot remove base shader!";
+		var prev : hxsl.ShaderList = null;
+		var hd = baseShaderList;
+		while ( hd != baseShaderListTail.next ) {
+			if ( hd.s == shader ) {
+				prev?.next = hd.next;
```
ddmills, line 336: getting an exception on this line when compiling 
image

h2d/RenderContext.hx
```diff
@@ -781,7 +830,7 @@ class RenderContext extends h3d.impl.RenderContext {
 			flush();
 		var shaderChanged = needInitShaders, paramsChanged = false;
 		var objShaders = obj.shaders;
-		var curShaders = currentShaders.next;
+		var curShaders = currentShadersTail.next;
```
ddmills, line 833: think this needs a null check

```suggestion
		var curShaders = currentShadersTail?.next;
```

h2d/RenderContext.hx
ddmills:
```suggestion
				if (prev != null) {
					prev.next = hd.next;
				}

```

ncannasse:
Shouldn't we have instead something like push/pop mechanism so we can use it during rendering ?

h2d/RenderContext.hx
Yanrishatum: No, because it's never supposed to be null. Tail points at the last global shader (`tail.next` points at object-specific shaders), and you should not be able to remove base shader from the global shader list, thus tail is never null.

h2d/RenderContext.hx
Yanrishatum: Not sure if that's newer Haxe or older Haxe not allowing such assignments with null syntax. Changed to `if`.

h2d/RenderContext.hx
ddmills: 🤷 idk i had to change this to make it work locally

Yanrishatum: Can you provide an repro sample?

PR #1267, opened 2025-01-23
Add hxd.net.RemoteConsole for local communication

yuxiaomao:
A basic string-based local communication with TCP, allow typically Hide-game communication


yuxiaomao:
Nicolas prefer have it in hide. Will move soon

PR #1268, opened 2025-01-27
Fix of nullable operator in ShaderCache.hx

p4tat0r:
It fixes the error message `home/user/haxelib/heaps/2,0,0/h3d/impl/ShaderCache.hx:14: characters 33-34 : Unexpected ?` when following the heaps tutorial https://heaps.io/documentation/hello-world.html  when building. Apparently haxe does not have the `??` operator like in php

### DIFF
--- a/h3d/impl/ShaderCache.hx
+++ b/h3d/impl/ShaderCache.hx
@@ -11,7 +11,7 @@ class ShaderCache {
 
 	public function new( file : String, ?outputFile : String ) {
 		this.file = file;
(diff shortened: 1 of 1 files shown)


ncannasse:
It's not a question of ?? but of haxe version. Which haxe version do you have ?

p4tat0r:
Thanks for your quick reply !  
I have the 4.2.5, which is not the last, but the one installed through apt.   
If this is due to my version not being the 4.3.6, I can close the PR.

PR #1270, opened 2025-02-03
Avoid conditionals in SDF shader

lbergman:
Have been looking in to an issue with WebGL context sometimes crashing on old Adreno GPUs when compiling shaders. 
Managed to resolve it by disabling mipmapping that was unnecessarily enabled on our text, but while investigating the issue I found that removing conditionals from the SignedDistanceField shader significantly reduced the crash frequency.

### DIFF
--- a/h3d/shader/SignedDistanceField.hx
+++ b/h3d/shader/SignedDistanceField.hx
@@ -29,13 +29,15 @@ class SignedDistanceField extends hxsl.Shader {
 
 		function fragment() {
 			var textureSample : Vec4 = textureColor;
-			var distance : Float;
-
-			distance = if (channel == 0) textureSample.r;
-				else if (channel == 1) textureSample.g;
-				else if (channel == 2) textureSample.b;
-				else if (channel == 3) textureSample.a;
-				else median(textureSample.r, textureSample.g, textureSample.b);
+			var distances = [
+				textureSample.r,
+				textureSample.g,
+				textureSample.b,
+				textureSample.a,
+				median(textureSample.r, textureSample.g, textureSample.b),
+			];
+			var ch = int(clamp(float(channel),0,4));
+			var distance = distances[ch];
 
 			var smoothVal = autoSmoothing ? abs(fwidth(distance) * 0.5) : smoothing;
 			textureColor = vec4(1.0, 1.0, 1.0, smoothstep(alphaCutoff - smoothVal, alphaCutoff + smoothVal, distance));


Yanrishatum:
Those `if` are supposed to be optimized out by HXSL compiler, as `channel` is a `@const`. Const variables will produce unique shader version based on its value, and all conditionals resolved. Enable shader dump and check for yourself, it should generate just `distance = textureSample.r` (or `median` calc)

lbergman:
Thanks, that is interesting, I will have a look and see if there is any difference in compiled shader.
Could just been luck I guess, but these are results of my test runs with and without this change: (☠️=crash, ❌=dead gl context, needs browser restart)

Without removed conditionals
☠️☠️✅✅✅☠️☠️☠️✅✅☠️
✅☠️✅✅☠️☠️❌☠️✅☠️✅✅✅✅☠️✅✅
With removed conditionals
✅✅☠️✅✅✅✅✅☠️☠️
✅✅✅✅✅☠️✅✅✅✅✅✅✅✅✅✅✅

lbergman:
To clarify, when it crashes it is due to failing to compile Base2d shader, see error below. Not really familiar with how the hxsl compilation works with imports and combining shaders, but since it fails at the compile stage, could it not be some issue with the conditionals that happens during compilation that would not be obvious by dumping the generated shader code?
```
"An error occurred compiling the shaders: 

1	#version 300 es
2	precision highp float;
3	in vec2 position;
4	in vec2 uv;
5	in vec4 color;
6	vec4 position_2;
7	vec4 spritePosition;
8	vec4 absolutePosition;
9	out vec4 pixelColor_varying;
10	out vec2 calculatedUV_varying;
11	vec4 outputPosition;
12	uniform vec4 vertexParams[6];
13	
14	vec3 tmp;
15	
16	void main(void) {
17		spritePosition = vec4(position, vertexParams[0].x, 1.);
18		absolutePosition.x = dot(vec3(spritePosition.xy, 1.), vertexParams[0].yzw);
19		absolutePosition.y = dot(vec3(spritePosition.xy, 1.), vertexParams[1].xyz);
20		absolutePosition.zw = spritePosition.zw;
21		tmp = vec3(absolutePosition.xy, 1.);
22		tmp = vec3(dot(tmp, vertexParams[2].xyz), dot(tmp, vertexParams[3].xyz), 1.);
23		outputPosition = vec4(dot(tmp, vertexParams[4].xyz), dot(tmp, vertexParams[5].xyz), absolutePosition.zw);
24		position_2 = outputPosition;
25		gl_Position = position_2;
26		calculatedUV_varying = uv;
27		pixelColor_varying = color;
28		gl_Position.z += gl_Position.z - gl_Position.w;
29	}
30	
31	"
```

TothBenoit:
In most cases, a crash without further information during shader compilation is due to a context loss by the driver and is very likely caused by a WebGL driver bug.
The HXSL compiler does not interact with WebGL, there are no hidden operations that could mess with the drivers.

lbergman:
Yes, it seems like the drivers are buggy since it is only reproduceable on some old specific versions of Adreno. And results of my tests was from when I had other code that actually triggered the bug.
When not doing mipmapping on an SDF there is no difference with this change since there are no crashes, and I'm fine with closing it.

Still, I do find it curious that it so clearly seemed to affect crash frequency on multiple sessions. I tried running with the `-D shader_debug_dump` but not getting any dump file, so have not been able to figure out what could be going on.

PR #1274, opened 2025-03-21
Fix preceding char getting cut off during word wrap and add option to disable word wrapping

miguelblix99:
This will solve the issue where a word gets broken into a new line and the previous char on the new line gets skipped. There is also now the option to disable mid word splitting.

### DIFF
--- a/h2d/Text.hx
+++ b/h2d/Text.hx
@@ -385,7 +385,7 @@ class Text extends Drawable {
 			else if( (x + esize + letterSpacing) - wLastSep > maxWidth && wordWrap ) {
 				newline = true;
 				lines.push(text.substr(restPos, i - restPos));
-				restPos = i + 1;
+				restPos = i;
 			}
 			if( e != null && cc != '\n'.code )
 				x += esize + letterSpacing;
@@ -447,7 +447,7 @@ class Text extends Drawable {
 			var cc = StringTools.fastCodeAt(t, i);
 			var e = font.getChar(cc);
 			var offs = e.getKerningOffset(prevChar);
-			var esize = e.width + offs;
+			var esize = e.t.dx + e.width + offs;
 			// if the next word goes past the max width, change it into a newline
 
 			if( cc == '\n'.code ) {
--- a/h2d/Text.hx
+++ b/h2d/Text.hx
@@ -382,10 +382,11 @@ class Text extends Drawable {
 				}
 				else wLastSep = size;
 			}
-			else if( (x + esize + letterSpacing) - wLastSep > maxWidth && wordWrap ) {
+			else if( (x + esize + letterSpacing) - wLastSep > (maxWidth-esize) && wordWrap ) {
 				newline = true;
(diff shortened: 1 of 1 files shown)


h2d/Text.hx
```diff
@@ -371,10 +382,11 @@ class Text extends Drawable {
 				}
 				else wLastSep = size;
 			}
-			else if( (x + esize + letterSpacing) - wLastSep > maxWidth ) {
+			else if( (x + esize + letterSpacing) - wLastSep > (maxWidth-esize) && wordWrap ) {
```
Yanrishatum, line 385: Probably should reverse the order of checks, and put `wordWrap` first, as first check becomes redundant when it's off.

h2d/Text.hx
```diff
@@ -371,10 +382,11 @@ class Text extends Drawable {
 				}
 				else wLastSep = size;
 			}
-			else if( (x + esize + letterSpacing) - wLastSep > maxWidth ) {
+			else if( (x + esize + letterSpacing) - wLastSep > (maxWidth-esize) && wordWrap ) {
 				newline = true;
 				lines.push(text.substr(restPos, i - restPos));
-				restPos = i + 1;
+				restPos = i;
+				x -= esize + letterSpacing;
```
Yanrishatum, line 389: Iirc it was `i+1` because word wrap was triggering on spaces, and during wrap we want to skip the space character. There should be a `charset.isSpace` check to determine if character should be skipped or not.

miguelblix99: I'm not sure i follow entirely as this is still handled in the previous condition, which means that no spaces will be included in the lines array during a linebreak as stated here:
```
if( lineBreak && (size > maxWidth || (!breakFound && size + afterData > maxWidth)) ) {
					newline = true;
					if( font.charset.isSpace(cc) ){
						lines.push(text.substr(restPos, i - restPos));
						e = null;
					}else{
						lines.push(text.substr(restPos, i + 1 - restPos));
					}
					restPos = i + 1;
				}
```

What i modified is part of the else block where the character is not a space nor any other break character.

kLabz: I think you mentioned the wrong `if` part, but your argument still applies. This is the `else` of `if( font.charset.isBreakChar(cc) && !isComplement ) {`, where `isBreakChar` checks for `isSpace` (and `isComplement` is `false` for space).

Could we go ahead with this PR? The issue is a bit ridiculous.. :/

h2d/Text.hx
```diff
@@ -371,10 +382,11 @@ class Text extends Drawable {
 				}
 				else wLastSep = size;
 			}
-			else if( (x + esize + letterSpacing) - wLastSep > maxWidth ) {
+			else if( wordWrap && (x + esize + letterSpacing) - wLastSep > (maxWidth-esize) ) {
```
kLabz, line 385: Why the `-esize` change, though? I'm now getting letters thrown to next line if the character ends exactly at max width

PR #1284, opened 2025-04-19
feat: Make Tile return self on self-modifying functions for easier chaining

Yanrishatum:
Makes the following h2d.Tile methods return self, to simplify chain calls:  
`setCenterRatio`, `flipX`, `flipY`, `setPosition`, `setSize`, `scaleToSize`, `scrollDiscrete`.

Reasoning: Fairly often I need to create a Bitmap with particular Tile configuration (most often centered), leading to `new Bitmap(Res.myTile.toTile().center())`, however it's very limiting (only centering) and also creates unnecessary Tile clone. This change allows for easier chain-configuration of such one-off tile loads, i.e. `Res.myTile.toTile().scaleToSize(32, 32).setCenterRatio(0, 0.5).flipX()`.

### DIFF
--- a/h2d/Tile.hx
+++ b/h2d/Tile.hx
@@ -159,7 +159,7 @@ class Tile {
 	}
 
 	/**
-		Returns a new Tile with shifting origin point (`dx` and `dy`) to the tile center.
+		Returns a **new Tile** with shifting origin point (`dx` and `dy`) to the tile center.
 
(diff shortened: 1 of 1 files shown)


ncannasse:
We don't do that because in Tile, returning a tile value means it's a modified copy and not a self reference. So it's not correct to have methods that are doing side effects but also returning this, it creates confusion.

PR #1285, opened 2025-05-05
Add hxd.tools.SceneInspector: domkit-based tool

yuxiaomao:
Allow inspect the hierarchy of s3d (and s2d) in a `hxd.App`, and display some useful information


yuxiaomao:
Close and is not planned for heaps (maybe in hide after stabilization)

PR #1290, opened 2025-07-16
Allow to set sample/bitrate when converting wav to mp3/ogg

tong:
Although resampling to 44100hz by default might be debatable, it remains the default behavior for backward compatibility.

### DIFF
--- a/hxd/fs/Convert.hx
+++ b/hxd/fs/Convert.hx
@@ -249,21 +249,19 @@ class ConvertWAV2MP3 extends Convert {
 	}
 
 	override function convert() {
-        var args = ["--silent"];
-        var sampleRate = 44100;
-        var bitRate = 128;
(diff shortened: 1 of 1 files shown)


yuxiaomao:
Please use tab instead of spaces for indentation, and only concat args if necessary (bitrate) so the previous behavior remains unchanged.

tong:
128 already was the default bitrate used for both converters, since:
* lame: the `-h` flag was set
* oggenc: by default, no flag was set

anyhow, i've changed it so it only gets set when specified.
the mp3 converter also sets the `-h` flag now if no custom bitrate is set (as before).

yuxiaomao:
Thank you!
I'm not familiar with these tools (and I didn't successfully find that `-h` implies a default 128 bitrate), it's easier for me to review if old behavior is kept.

PR #1292, opened 2025-07-18
sdl/dx: fix missing mouse position for KeyDown/KeyUp events

tong:
Ensures mouse position is included in EKeyDown/Up events so that h2d.Interactive focus checks work correctly for keyboard events.

### DIFF
--- a/hxd/Window.hl.hx
+++ b/hxd/Window.hl.hx
@@ -515,14 +515,14 @@ class Window {
 
 		#elseif hldx
 		case KeyDown:
-			eh = new Event(EKeyDown);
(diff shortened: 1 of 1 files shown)


tong:
it's a nasty bug.
you can test it with the interactive2D sample.
if you add a onKeyDown handler to h2d.Interactive the event should fire when it has focus. but since the mouse position isn't added to the EKeyDown/EKeyUp events it never does since skipped the focus check here: https://github.com/HeapsIO/heaps/blob/0aaa335fc16f9306803d10c70fb47e1cfae6ed8b/h2d/Scene.hx#L547-L555

iv'e added it for dx, but cannot test it.

PR #1295, opened 2025-08-21
Fix Quat.slerp using Godot Engine implementation

Zylann:
Replaces the old `h3d.Quat.slerp` with a new implementation ported from [Godot Engine](https://github.com/godotengine/godot/blob/d5ad0556a2c04c50694b5c04dc1b3bf03ecd7113/core/math/quaternion.cpp#L101)

May fix https://github.com/HeapsIO/heaps/issues/1088

### DIFF
--- a/h3d/Quat.hx
+++ b/h3d/Quat.hx
@@ -43,7 +43,7 @@ class Quat {
 		this.w = q.w;
 	}
 
-	public function clone() {
+	public inline function clone() {
 		return new Quat(x, y, z, w);
 	}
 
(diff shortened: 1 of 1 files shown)


Zylann:
I made a test project that shows slerp between different bases:
[Quat.zip](https://github.com/user-attachments/files/21941731/Quat.zip)

Here is an example, first showing the old slerp, and then switching to the new slerp:
https://github.com/user-attachments/assets/01d5c79f-afe5-4401-9350-7935cf2a12ed

Here with 100 of these cases, again showing the old slerp, then switching to the new:
https://github.com/user-attachments/assets/72d726d8-db2f-47c2-b4fd-8eadee4b2ffd

Regarding axis difference, I think for this function it doesn't matter. I haven't encountered an issue of that kind when using it.

PR #1296, opened 2025-08-27
Fix Switch build failed 

mrpierrot:
On Console (Test on Switch), call of sys.io.Process trigger that error : `error : undefined symbol: hl_process_stdin_close`
It's fine because we don't need it on console. But in the SourceLoader.hx there is a call of it keeped in console. 
I place a test on preprocessor usesys toi avoid that.


yuxiaomao:
Thank you!

PR #1313, opened 2025-10-28
add anti aliasing option

mc-saw:
Normally to reduce aliasing in games mip maps are used, but mip maps increases the amount of texture memory used with around 50%. Another way to get the same effect as mip maps is to sample the texture 4 times and take an average of the color value. This solution works good on web devices where we would like to minimize the used memory. To the users who want this type of anti aliasing they can set the compile flag `-D anti_aliasing` in their `build.hxml` and also set the runtime flag `globals.set("global.antiAliasing", 1);`.

### DIFF
--- a/h2d/RenderContext.hx
+++ b/h2d/RenderContext.hx
@@ -151,6 +151,9 @@ class RenderContext extends h3d.impl.RenderContext {
 		cameraStackIndex = 0;
 		filterStack = [];
 		filterStackIndex = 0;
+		#if anti_aliasing
+		globals.set("global.antiAliasing", 0);
+		#end
 	}
 
 	override function dispose() {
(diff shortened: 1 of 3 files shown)


lbergman:
The reason for putting it at texture level is that this becomes a problem with all textures with some browser window sizes on some resolutions. For example if I open a game on my 42" 4K monitor with device pixel ratio of 1:1, and scale down browser to only use a part of the screen, aliasing becomes really bad.
On devices with a high device pixel ratio, like mobile devices or many desktop screens with OS Scaling, it is not an issue, so we like to enable it in runtime when it is actually needed.

I agree that ideally it would be a separate shader, but not sure if there is a nice way to add a shader to all textures?

PR #1318, opened 2025-11-24
[Flow] Fix line height of first line in multiline flow

kLabz:
The issue can be easily reproduced with this sample:
```haxe
import h2d.Tile;
import h2d.Bitmap;
import h2d.Flow;

class Main extends hxd.App {
	static public function main() {
		new Main();
	}

	override function init() {
		var flow = new Flow(s2d);
		flow.maxWidth = 200;
		flow.multiline = true;
		flow.debug = true;

		new Bitmap(Tile.fromColor(0xFF00FF, 200, 10), flow);
		new Bitmap(Tile.fromColor(0x00FFFF, 20, 40), flow);
		new Bitmap(Tile.fromColor(0xFFFF00, 10, 20), flow);

		var flow = new Flow(s2d);
		flow.maxWidth = 200;
		flow.multiline = true;
		flow.verticalAlign = Top;
		flow.x = 250;
		flow.debug = true;

		new Bitmap(Tile.fromColor(0x00FFFF, 200, 10), flow);
		new Bitmap(Tile.fromColor(0x00FFFF, 20, 40), flow);
		new Bitmap(Tile.fromColor(0xFFFF00, 10, 20), flow);
	}
}
```

Resetting `maxLineHeight` to `0` before positioning nodes makes sure previously calculated `maxLineHeight` isn't being used for first line when it should not.

Current heaps:

With this patch:

_Tested with shiroTools `73faca1146`_

### DIFF
--- a/h2d/Flow.hx
+++ b/h2d/Flow.hx
@@ -1426,6 +1426,7 @@ class Flow extends Object {
 			});
 
 			// position all not absolute nodes
+			maxLineHeight = 0;
 			forChildren(function(i, p, c) {
 				if( p.autoSizeWidth != null || p.autoSizeHeight != null )
 					calcSize(p, c);
--- a/h2d/Flow.hx
+++ b/h2d/Flow.hx
@@ -1625,6 +1625,7 @@ class Flow extends Object {
 			});
 
 			// position all not absolute nodes
+			maxColWidth = 0;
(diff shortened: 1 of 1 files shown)


ncannasse:
Shouldn't it also be done for columns in other align ?

kLabz:
Can we merge this?

PR #1335, opened 2026-02-08
Add Opus audio format support

dtwotwo:
Add Opus decoding support for HashLink target.
Requires: approved HashLink pull request - [HashLink#875](https://github.com/HaxeFoundation/hashlink/pull/875)


dtwotwo:
Use this library for heaps to get Opus support: https://github.com/gimmesoda/hlopus.
It is based on my pull request code and works independently, without modifying HashLink or Heaps.
It would be better to use it as a separate library instead of this one.

PR #1350, opened 2026-05-27
Streamline / DLAA

julesvannier:
Heaps native folder, hlheaps sln to add native feature to heaps without relying on hashlink.
Streamline API integration in Haxe
Heaps DLSS API
DLAA working, DLSS not tested yet


ncannasse:
I think we should keep the native part separated, we actually have https://github.com/HeapsIO/hlheaps that would be a better repo.

ncannasse:
Also the haxe code matching the native code should be in hlheaps, only the heaps "api user" code should be in heaps, with #if hlheaps .. #else throw "Requires -lib hlheaps"; #end

PR #1354, opened 2026-06-25
Fix segments contains

Kubuzer:
This Pull Request implements the missing non-convex (concave) shape support in `h2d.col.Segments.containsPoint` and introduces a headless unit testing suite.

### Changes:
* **`h2d.col.Segments`**: Replaced the placeholder `throw "TODO"` exception in `containsPoint` (when `isConvex = false`) with a ray-casting (winding number) algorithm to support concave polygons.
* **Testing Suite**: Created a lightweight testing framework under `test/` running on the Haxe compiler interpreter (`--interp`) to ensure headless tests can run on any platform.
  * Added math tests for `hxd.Math` (`clamp`, `nextPOT`, `isPOT`, `lerp`, etc.).
  * Added collision tests for `h2d.col.Segments` verifying both convex and concave polygon containment.
  * Added regression test for `h3d.impl.ShaderCache` ensuring shaders are cached in memory when `allowSave = false`.

### DIFF
--- a/h2d/col/Segments.hx
+++ b/h2d/col/Segments.hx
@@ -40,7 +40,20 @@ abstract Segments(Array<Segment>) from Array<Segment> to Array<Segment> {
 				if( s.side(p) < 0 )
 					return false;
 		} else {
-			throw "TODO";
+			// Ray-casting (winding number) algorithm for arbitrary polygons.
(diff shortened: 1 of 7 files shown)


yuxiaomao:
I think we'll not accept unit test files for the moment, please remove it from this PR

PR #1357, opened 2026-07-10
Fix nintendo controllers vs sdl3

kLabz:
Since the SDL3 migration, nintendo controllers have A<->B and X<->Y buttons swapped


nspitko:
Ok after thinking about it for a bit, this is probably the right thing to do here for now. There's likely a wider discussion to be had around how pads are handled in heaps, but I think that should be a wider refactor around the new sdl3 apis, and for now we should just focus on maintaining the old code behavior.

PR #1359, opened 2026-07-10
Fix symlinks in LocalFileSystem

kLabz:
On my project, `res/data.cdb` (and some other files/directories) is a symlink to somewhere else, and it broke with https://github.com/HeapsIO/heaps/commit/da90afff


yuxiaomao:
Approved by @Speedphoenix

